Issue #11, opened 2025-05-27
Update docker compose to a working state


ways:
... and a suitable way to run the project.

I will supply a series of pull requests to fix this issue.

PR #14, opened 2025-05-27
Update docker-compose.yml and frontend/Dockerfile. 

ways:
Note that frontend/Dockerfile contains debug tools not meant for production.

Related to https://github.com/ecmwf/forecast-in-a-box/issues/11

But I'm unable to get frontend to "see" backend and cascade. It works from cli:

```
> docker compose exec frontend bash
root@249d09954aee:/app/frontend# curl backend:8000
{"detail":"Not Found"}root@249d09954aee:/app/frontend# 
root@249d09954aee:/app/frontend# curl cascade:8067
curl: (1) Received HTTP/0.9 when not allowed
```

```
> docker compose exec backend bash
root@2112f5eac7f2:/app/backend# curl cascade:8067
curl: (1) Received HTTP/0.9 when not allowed
```

I'm not a frontend dev, so don't know how to debug the node app.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,37 +1,32 @@
 name: FIAB
-
 services:
+
   frontend:
     container_name: fiab-frontend
     build:
       context: ./frontend
       dockerfile: Dockerfile
-    image: eccr.ecmwf.int/forecast-in-a-box/fiab-frontend:latest
+
     ports:
-      - "127.0.0.1:3000:3000"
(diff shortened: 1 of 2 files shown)


HCookie:
I'll take a look at the node connection issue

HCookie:
Because of how the web server partly runs on the client, intercontainer communication doesn't seem possible, but using nginx to forward api calls is fine.

PR #38, opened 2025-06-12
[backend] Add unit tests

HCookie:


backend/tests/unit/models/test_model.py
```diff
@@ -0,0 +1,98 @@
...
+    assert qube.span("param") == ["10u", "10v", "2t", "q", "tcc", "tp"], "Qube 'param' axis should match expected values"
+    assert "levtype" in qube.axes(), "Qube should have 'levtype' axis"
+    assert qube.span("levtype") == ["pl", "sfc"], "Qube 'levtype' axis should match expected values"
+    assert "levelist" in qube.axes(), "Qube should have 'levelist' axis"
+    assert qube.span("levelist") == ["850"], "Qube 'levelist' axis should match expected values"
+    assert "frequency" in qube.axes(), "Qube should have 'frequency' axis"
+    assert qube.span("frequency") == [datetime.timedelta(hours=6)], "Qube 'frequency' axis should match expected values"
+
+
+@fake_checkpoints
```
tmi, line 76: what is the `fake_checkpoints` doing? Patching something inside anemoi.inference? That could be unfortunate if we wanted to convert this test into an integration one (and I would want that) -- because patches don't carry over across process boundaries

HCookie: It does, it's currently the only way to test inference

backend/tests/unit/models/test_model.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+@pytest.fixture
+@fake_checkpoints
+def mock_checkpoint(mock_checkpoint_path) -> Checkpoint:
+    return Checkpoint(mock_checkpoint_path)
+
+
+@pytest.fixture
+def test_model(mock_checkpoint_path, mock_checkpoint) -> Model:
```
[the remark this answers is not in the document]
HCookie: I liked having fixtures to be able to reuse and keep things seperate...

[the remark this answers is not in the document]
tmi: yes, but why via fixture? `(Path(__file__).parent / "../checkpoints/simple.yaml").absolute()` is a constant, not a fixture :) Similarly TestModel below is either a factory or a singleton...

This is one of my long standing rants, for some reason many python people do regular code but once they enter `tests` they are like "oh wait everything is a fixture now". Fixtures have their role when you need to mix factories, singletons, scope initializers and the like in a way that would be hassle without accessing pytest setup/teardown internals. But that doesn't mean every reusable piece of code in tests should be a fixture :)

this is not a request to change, the code works as is -- it's just a general quality improvement suggestion

backend/tests/unit/models/test_model.py
tmi: but what is it doing? Why don't we have a genuine tiny model which runs fast?

HCookie: This handles the metadata initialisation, and pretends to be a model, we could switch it to a genuine model, but would require storing it as well

PR #39, opened 2025-06-12
[backend] ci: Add workflows

HCookie:


backend/pyproject.toml
```diff
@@ -29,9 +29,9 @@ dependencies = [
 	"qubed==0.1.12",
 	"earthkit-workflows-anemoi>=0.3.1",
 	"pydantic-settings",
+	"pproc@git+https://github.com/ecmwf/pproc",
+	"earthkit-workflows-pproc@git+https://github.com/ecmwf/earthkit-workflows-pproc",
```
tmi, line 33: are we dropping wheels for those packages on purpose?

HCookie: Don't think they had wheels when I made this PR

PR #43, opened 2025-06-19
persistence to sqlite

tmi:
Refactorings for the /job and /graph api impls:
1/ removing lru cache on job results, due to redundancy
2/ removing background tasks from job submit due to not being needed
3/ unifying id and job_id
4/ not sorting the job progresses due to order not being json guaranteed
5/ persisting JobRecord in sqlite instead of mongodb
6/ reducing the total amount of db queries by a substantial amount (way over 50%)
7/ add integration test coverage for /delete and /flush

note integration tests now pass without any mongodb instance -- but thats in part due to there being no coverage of the /model api :), which still uses mongodb -- i'll get rid of that in the next PR

### DIFF
--- a/backend/forecastbox/api/routers/job.py
+++ b/backend/forecastbox/api/routers/job.py
@@ -93,7 +93,7 @@ async def update_and_get_progress(job_id: JobId) -> JobProgressResponse:
         response: api.JobProgressResponse = client.request_response(
             api.JobProgressRequest(job_ids=[job_id]), f"{config.cascade.cascade_url}"
         )  # type: ignore
-    except TimeoutError | KeyError | Exception as e:
+    except (TimeoutError, KeyError, Exception) as e:
         status = {TimeoutError: "timeouted", KeyError: "invalid", Exception: "errored"}[e.__class__]
         error_on_request = repr(e)
     if response.error:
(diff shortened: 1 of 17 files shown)


backend/forecastbox/api/routers/job.py
```diff
@@ -73,138 +73,99 @@ class JobProgressResponses:
...
-    """Validate the job ID by checking if it exists in the database."""
-    collection = db.get_collection("job_records")
-
-    record = collection.find_one({"job_id": job_id})
-    if record:
-        return job_id
-    raise HTTPException(status_code=404, detail=f"Job {job_id} not found in the database.")
+    """Validate the job ID."""
+    # NOTE we could query the db here, but since next step is a conditional db retrieval anyway, this extra query makes low sense
+    return job_id
```
HCookie, line 78: By not validating here, are you not pushing the requirements to throw a nice error on all downstream functions rather then centralising it?

HCookie (approved):

backend/forecastbox/api/routers/job.py
tmi: they have been there anyway :) To avoid that, you would have to return the whole entity from the validation. You have no guarantee that if the entity exists at the start of the request, then it will exist for the whole duration of the request. So _each_ query into the database must be guarded against None-return

PR #50, opened 2025-06-26
[backend] fix job error handling and gateway start

tmi:
Fixes gateway logging after restart -- for some reason, new process doesnt reflect os environ changes, so I switch over to regular parameter passing

Extends the job status test to include gateway restart

Fixes job status endpoint to handle multiple cases better -- we rely on the DB state unless running/submitted. We are more careful in persisting DB updates, eg timeout should not cause transition from running.

Not found in gateway is now "invalid" with a better error message

Lastly, updates cascade version to the one without improper-gateway shutdown bug -- this caused the tests to spam into the console long after pytest finished

### DIFF
--- a/backend/forecastbox/api/routers/gateway.py
+++ b/backend/forecastbox/api/routers/gateway.py
@@ -88,7 +88,7 @@ async def start_gateway() -> None:
 
     logs_directory = TemporaryDirectory()
     log_path = GatewayProcess.log_path(logs_directory)
-    # NOTE for some reason changes to os.environ were *not* visible by the child process!
+    # TODO for some reason changes to os.environ were *not* visible by the child process! Investigate and re-enable:
     # export_recursive(config.model_dump(), config.model_config["env_nested_delimiter"], config.model_config["env_prefix"])
     process = Process(target=launch_cascade, args=(log_path,))
     process.start()
--- a/backend/forecastbox/api/routers/job.py
+++ b/backend/forecastbox/api/routers/job.py
@@ -98,7 +98,11 @@ async def update_and_get_progress(job_id: JobId) -> JobProgressResponse:
             )  # type: ignore
         except TimeoutError:
             # NOTE we dont update db because the job may still be running
-            return JobProgressResponse(progress="0.00", created_at=created_at, status="timeout")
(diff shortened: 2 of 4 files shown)


backend/forecastbox/api/routers/gateway.py
```diff
@@ -87,12 +87,13 @@ async def start_gateway() -> None:
             gateway = None
 
     logs_directory = TemporaryDirectory()
-    config.cascade.log_path = GatewayProcess.log_path(logs_directory)
-    export_recursive(config.model_dump(), config.model_config["env_nested_delimiter"], config.model_config["env_prefix"])
-    process = Process(target=launch_cascade)
+    log_path = GatewayProcess.log_path(logs_directory)
+    # NOTE for some reason changes to os.environ were *not* visible by the child process!
+    # export_recursive(config.model_dump(), config.model_config["env_nested_delimiter"], config.model_config["env_prefix"])
```
HCookie, line 92: Can this be removed?

backend/forecastbox/api/routers/job.py
```diff
@@ -85,51 +85,48 @@ def validate_dataset_id(dataset_id: str) -> str:
...
+
+    # NOTE the submitted status currently shouldnt happen due to linearization of submit-insert
+    if job.status in ("running", "submitted"):
+        try:
+            response: api.JobProgressResponse = client.request_response(
+                api.JobProgressRequest(job_ids=[job_id]), f"{config.cascade.cascade_url}"
+            )  # type: ignore
+        except TimeoutError:
+            # NOTE we dont update db because the job may still be running
+            return JobProgressResponse(progress="0.00", created_at=created_at, status="timeout")
```
HCookie, line 101: No longer handling `KeyError`'s that get raised? or did that change?

tmi: thats the case `response.error.startswith("KeyError"):` below. The gateway itself raises KeyError, catches it, and stuffs it into response. The `client.request_response` itself does not raise it

huh, looking at the code, I should replace the TimeoutError with ValueError, or even better change the gateway code to actually raise Timeout and not wrap it in ValueError :/

PR #57, opened 2025-07-02
[ci] add basic ci action

tmi:
Just to allow me testing macos runs. Proper checks, triggers, full matrix, etc will come later


tmi:
cc @HCookie ; I'm merging to have github allow running this action

PR #65, opened 2025-07-09
[cicd] minor fixes

tmi:
Autogenerate manifest.in to make sure everything gets included (other solutions I came up with required eg committing all generated files to git, or explicitly listing all extensions)
Fix testpypi publish
Add lower bound to fiab.sh installer to circumvent our personal caches


tmi:
cc @HCookie nothing interesting here, merging right away but feel free to comment as always

PR #75, opened 2025-07-11
[backend] /api/v1/job/jobid/logs endpoint

tmi:
adds a new endpoint that returns a zip file with all the logs for the particular job and other information (gateway, db state, etc)

intended for troubleshooting / remote support


tmi:
> This will only work if the gateway is launched from the entrypoint, if launched as a seperate process it will mislead the user.

true, but we dont expect to launch gateway from elsewhere in fiab; its the same with `/gateway/logs`

HCookie (approved):

PR #79, opened 2025-07-25
feat: Add react-jsonschema-form to product configuration

HCookie:
### Description
Uses react-jsonschema-form to define product configuration
- Moves product export into graph

TODO: Use in model configuration


.pre-commit-config.yaml
```diff
@@ -1,12 +1,84 @@
...
+-   repo: https://github.com/pre-commit/pygrep-hooks
+    rev: v1.10.0  # Use the ref you want to point at
+    hooks:
+    -   id: python-use-type-annotations # Check for missing type annotations
+    -   id: python-check-blanket-noqa # Check for # noqa: all
+    -   id: python-no-log-warn # Check for log.warn
+- repo: https://github.com/psf/black-pre-commit-mirror
+  rev: 25.1.0
+  hooks:
+  - id: black
```
tmi, line 32: why would we have black when we have ruff? Lets remove black

backend/forecastbox/api/routers/product.py
```diff
@@ -59,38 +63,56 @@ def select_from_params(available_spec: Qube, params: dict[str, Any]) -> Qube:
 
     return available_spec
 
+def _sort_values(values: list[Any]) -> list[Any]:
+    if all(str(x).isdigit() for x in values):
+        values = list(map(float, sorted(values, key=float)))
+        return values
+    values = list(sorted(values, key=lambda x: str(x).lower()))
+    return values
```
tmi, line 71: puzzles me a bit -- this sort a numerical list (say ["12", "9"]) in numerical order (["9", "12"]), but if its mixed (["12", "9", "hello"]) all of sudden its reversed (ie "12" will come before "9")

dont you want some sort of https://stackoverflow.com/questions/2669059/how-to-sort-alpha-numeric-set-in-python to make it stable? The first solution there is nice

backend/forecastbox/api/routers/product.py
```diff
@@ -59,38 +63,56 @@ def select_from_params(available_spec: Qube, params: dict[str, Any]) -> Qube:
...
+    new_fields = OrderedDict()
+    for key in CONFIG_ORDER:
+        if key in fields:
+            new_fields[key] = fields[key]
+
+    for key in fields:
+        if key not in new_fields:
+            new_fields[key] = fields[key]
+
+    return new_fields
```
tmi, line 83: just be aware that dictionaries dont preserve order when serialized... list of tuples safer

backend/forecastbox/products/rjsf/jsonSchema.py
```diff
@@ -0,0 +1,233 @@
...
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+# A pydantic implementation of JSON Schema for React JSON Schema Form (RJSF).
+# This schema defines the structure of data as expected by RJSF based on the JSON Schema specification.
+# https://rjsf-team.github.io/react-jsonschema-form/docs/
+
+from typing import Any
+from typing import Dict
+from typing import List
```
tmi, line 16: I suggest replacing `Dict` and `List` with `dict` and `list`, we are on high enough python

PR #82, opened 2025-07-30
[backend] relax timeouts in bigscript, reduce logging

tmi:
1/ eliminate one very annoying logging kind
2/ relax timeouts for big script
3/ flush the progressing-downloads db on start


HCookie (approved):

havardf:
I still get timeouts on the model download part, although currently I am sitting on a slow network. Maybe just triple the max iterations? And also, print out the download progress info so it is easier to see what is going on.

tmi:
@havardf iterations for download trippled + logging added

@HCookie ftr, I've also added a fix to have the un-logged user behave like admin since you've reviewed

PR #86, opened 2025-08-06
[ci] bigtest opendata reading

tmi:
Switches bigtest to read opendata

Adds options to allow everything-to-stdout logging

Adds earthkit caching when launching in dev mode

Switches to cascade_platform mp ctx so that we `spawn` on macos -- which should be safer


HCookie (approved):
For now seems reasonable, I'll update the config of the run for bigtest in due time to allow more generic runs.

PR #109, opened 2025-08-19
[backend.docs] just adding troubleshooting faq

tmi:


tmi:
@HCookie jfyi. I'll clean it up later and turn into better docs, so far just collecting all the pleasantries we've encountered to make sure we don't forget (esp since our slack history is cut off)

PR #126, opened 2025-09-15
feat: [settings, rjsf] Update settings to use rjsf

HCookie:
### Description
Move to rjsf for admin settings

Also add config saving

### Contributor Declaration

### DIFF
--- a/backend/forecastbox/api/execution.py
+++ b/backend/forecastbox/api/execution.py
@@ -141,6 +141,4 @@ async def execute(spec: ExecutionSpecification, user: UserRead | None) -> Submit
         spec.model_dump_json(),
         json.dumps(list(map(lambda x: x.task, sinks))),
     )
-    if response.error:
-        raise Exception(response.error)
     return SubmitJobResponse(id=response.job_id)
--- a/backend/forecastbox/api/execution.py
+++ b/backend/forecastbox/api/execution.py
@@ -85,7 +85,7 @@ def _execute_cascade(spec: ExecutionSpecification) -> tuple[api.SubmitJobRespons
     try:
         job, job_envvars = execution_specification_to_cascade(spec)
     except Exception as e:
-        return api.SubmitJobResponse(job_id=None, error=str(e)), []
+        return api.SubmitJobResponse(job_id=None, error=repr(e)), []
 
     sinks = cascade_views.sinks(job)
     sinks = [s for s in sinks if not s.task.startswith("run_as_earthkit")]
--- a/backend/forecastbox/api/execution.py
+++ b/backend/forecastbox/api/execution.py
@@ -85,7 +85,7 @@ def _execute_cascade(spec: ExecutionSpecification) -> tuple[api.SubmitJobRespons
     try:
         job, job_envvars = execution_specification_to_cascade(spec)
     except Exception as e:
-        return api.SubmitJobResponse(job_id=None, error=repr(e)), []
+        return api.SubmitJobResponse(job_id=None, error=str(e)), []
 
     sinks = cascade_views.sinks(job)
     sinks = [s for s in sinks if not s.task.startswith("run_as_earthkit")]
@@ -141,4 +141,6 @@ async def execute(spec: ExecutionSpecification, user: UserRead | None) -> Submit
         spec.model_dump_json(),
         json.dumps(list(map(lambda x: x.task, sinks))),
     )
+    if response.error:
+        raise Exception(response.error)
     return SubmitJobResponse(id=response.job_id)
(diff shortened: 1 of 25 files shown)


backend/forecastbox/api/routers/admin.py
```diff
@@ -41,31 +42,48 @@ def get_admin_user(user: UserRead | None = Depends(current_active_user)) -> User
...
             setattr(old, key, val)
 
     try:
         update(config.api, settings.api)
+        update(config.product, settings.product)
         update(config.cascade, settings.cascade)
     except Exception as e:
         return HTMLResponse(content=str(e), status_code=500)
 
+    config.save_to_file()
```
tmi, line 85: whoa... where does this save to? And does it affect all places we want to affect? Note eg we serialize the config to pass to the gateway process, so shouldnt we restart the gateway as well?

backend/forecastbox/config.py
```diff
@@ -157,12 +166,53 @@ class FIABConfig(BaseSettings):
...
+        file_secret_settings: PydanticBaseSettingsSource,
+    ) -> tuple[PydanticBaseSettingsSource, ...]:
+        return env_settings, file_secret_settings, dotenv_settings, TomlConfigSettingsSource(settings_cls, fiab_home / "config.toml"), init_settings
+
+    def _get_toml(self, skip_secrets: bool = True, **k) -> str:
+        json_config = self.model_dump(mode="json", **k)
+
+        if skip_secrets:
+            (json_config.get('auth', {}).get('oidc', {}) or {}).pop('client_secret', None)
+            json_config.get('auth', {}).pop('jwt_secret', None)
```
tmi, line 193: we should instead recursively crawl and pop all values of type SecretStr... we have some recursive crawl already here, we can generalize

tmi: or we can have a custom child of SecretStr with overridden ser as in https://github.com/pydantic/pydantic/issues/6778, thats neat too

backend/forecastbox/config.py
```diff
@@ -177,3 +227,6 @@ def validate_runtime(config: FIABConfig) -> None:
 
 
 config = FIABConfig()
+config.save_example_to_file()
```
tmi, line 230: on every import of this module this would be triggered? So runs of mypy, pytest, etc, ... all would trigger a save to the possibly non-existent ~/.fiab? Thats sounds unhealthy

backend/forecastbox/config.py
```diff
@@ -157,12 +166,53 @@ class FIABConfig(BaseSettings):
...
+    def save_example_to_file(self) -> None:
+        """Save example configuration to toml file"""
+
+        config_path = fiab_home / "config.toml.example"
+        config_path.parent.mkdir(parents=True, exist_ok=True)
+        with open(config_path, 'w') as f:
+            f.write(self._get_toml())
+
+    def __del__(self) -> None:
+        self.save_to_file()
```
tmi, line 215: this also feels like a very heavy handed thing... could lead to some weird situations, like shutting down mid-initialization possibly destroys your config file

backend/forecastbox/config.py
```diff
@@ -28,6 +31,12 @@ class StatusMessage:
     # NOTE this class is here as this is a low place in hierarchy, and we dont want circular imports
     gateway_running = "running"
 
+class Secrets(SecretStr):
+    def __str__(self) -> str:
+        return self.get_secret_value()
```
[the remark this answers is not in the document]
HCookie: It seems as if even on serialisation the string is hidden, it can likely be removed, as other work has rendered this pointless. I just need to confirm the ordering of init args

backend/forecastbox/api/routers/admin.py
HCookie: Primarily these settings are used to control the products, so a restart of the gateway probably isn't needed

HCookie:
My git-foo failed, so maybe not....

backend/forecastbox/api/execution.py
```diff
@@ -85,7 +85,7 @@ def _execute_cascade(spec: ExecutionSpecification) -> tuple[api.SubmitJobRespons
     try:
         job, job_envvars = execution_specification_to_cascade(spec)
     except Exception as e:
-        return api.SubmitJobResponse(job_id=None, error=repr(e)), []
+        return api.SubmitJobResponse(job_id=None, error=str(e)), []
```
tmi, line 88: why is this being changed? `str(e)` is usually less informative than `repr(e)`, in particular the class gets lost

backend/forecastbox/models/globe.py
```diff
@@ -10,10 +10,12 @@
 
 from typing import Any
 
+from forecastbox.config import config
+
 from .metadata import ControlMetadata
 from .model import BaseForecastModel
 
-DEFAULT_GLOBAL_INPUT_SOURCE = "opendata"
+DEFAULT_GLOBAL_INPUT_SOURCE = config.product.default_input_source or 'opendata'
```
tmi, line 18: we dont need the `or` anymore, do we? Since the config field is non-nullablle and has a default

Issue #158, opened 2025-10-06
Improve pagination


tmi:
Improve job pagination by doing the logic in sql primarily -- limit and filtering in particular

Fix the job count field documentation

Introduce a consistent ordering based on created time

PR #159, opened 2025-10-06
[backend.jobs] improve pagination

tmi:
dont fetch all records from db, instead:
 - use offset and limit to get only the right number
 - filter via where to get only those with matching status

fix a bug where we didnt report the progress as known by the db in case of completed/failed jobs

NOTE: didnt test yet (and wont today), but can be already reviewed

closes https://github.com/ecmwf/forecast-in-a-box/issues/158


HCookie (approved):
Tested on the dev server, LGTM

PR #171, opened 2025-10-17
Fix typing and enable ty on ci

tmi:
The integration tests take quite long, I want to have something faster to catch basic issues

Lot of changes, mostly due to db interactions. But much less pain than I feared

I went with `ty`, happy to switch to any other

Additionally I fix CI -- no more runs on unsupported x86, and actually use the python version from the matrix :D

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        arch_type: [ARM64, X64]
+        arch_type: [ARM64]
         python_version: ["3.11", "3.12", "3.13"]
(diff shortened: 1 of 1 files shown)


backend/forecastbox/models/model.py
```diff
@@ -280,7 +280,7 @@ def form(self) -> FormDefinition:
                     title="Date",
                     description="The date for the forecast",
                 ),
-                uischema=UIStringField(widget="date", options={'yearsRange': [-20, 0]}),
+                uischema=UIStringField(widget="date", options={'yearsRange': [-20, 0]}), # type: ignore[unknown-argument] # TODO harrison `options` does really not seem to be declared!
```
tmi: @HCookie this is perhaps the only change here you may want to look into

PR #175, opened 2025-10-22
[docs] add arch and c4 diags

tmi:
Basic arch tech docs


HCookie (approved):
Amazing, thanks heaps for this

PR #181, opened 2025-10-28
[github] fix docker build yaml

tmi:
fixing yaml


tmi:
merging right away, no real change. But the original PR was so wrong that github hasnt registered the dispatch hook or something I guess

PR #189, opened 2025-10-31
Backend API to support new update functionality

tmi:
Just two new endpoints, `get /update` and `post /update?tag`, where the `get` gives how old is the current version, what is the current version, and what is the most recent version, and the `post` updates to the newest version / explicitly specified version

The update is just replacement of the pylock.toml file, so on the _next_ start of the app, a pip install will actually do the update

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,8 +19,11 @@ repos:
   - id: trailing-whitespace # Trailing whitespace checker
   - id: no-commit-to-branch # Prevent committing to main / master
   - id: check-added-large-files # Check for large files added to git
-    exclude: '.*uv.lock'
-    exclude: '.*pylock.toml'
+    exclude: |
+      (?x)(
+        .*uv.lock|
+        .*pylock.toml
+      )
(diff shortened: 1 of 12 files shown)


liefra:
The newly added Release Endpoint Signatures look good to me. Just wondering whether we need admin authentication for the GET endpoint `/release` ? We may want to display this release version to everyone on the frontend as part of a potential debugging workflow later on? Or maybe add it as another public endpoint to status? What do you think?

HCookie (approved):

liefra:
> what I can and will do is to add the current release number to the existing status endpoint as you say -- meaning it won't query github for the newest release, just return whats local -- so that the `/status` is fast. And include it in the `/{job_id}/logs` endpoint's output since thats what we'll use for the troubleshooting. Maybe I can include a whole `pip freeze` output in there just to be sure

+1 from my side. And yes, the newest local release will be sufficient for `/status` to avoid the network request here. I have no strong opinion on whether the `pip freeze` output is needed, but it doesn’t hurt either. I suppose it could prove useful for debugging purposes. Thanks.

Issue #193, opened 2025-11-07
Interloop+Frontend development setup


tmi:
We need to ensure Interloop developers have a working development flow

This includes:
1. [x] CI runs (current one is `pull_request` so won't trigger on fork-PRs, and uses ecmwf hosted runners... consider enabling ubuntu github free tier on forks + single python version)
2. [ ] Extension of testing to cover the frontend being buildable/servable

tmi:
@liefra thx, just wanted to make sure its correct not to close this Feature just yet :) But speaking of that, would there be anything else to it? I've left a `3. ...` in the feature description at the time of writing, but can't think of anything atm...

tmi:
@liefra I originally didnt want to because the item `Extension of testing to cover the frontend being buildable/servable` is actually not addressed :) But it does not fit well into a task called "Development setup", so I'll address it in the upcoming task "improve bigtest" -- we'll have some browser-automation-based test there

PR #198, opened 2025-11-13
[ci] extend ci runners

tmi:
for the self hosted runners, we add a guard to not execute on PRs targetting a fork -- they would timeout anyway

we add ubuntu-based action on github runners, this triggers on any PR

this addresse concern (1) from https://github.com/ecmwf/forecast-in-a-box/issues/193

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,12 +20,36 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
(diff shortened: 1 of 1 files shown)


tmi:
> Looks goods to me. Once we have frontend test included we can extend the ubuntu runner accordingly.

just ftr, I'd rather extend both actions, ie, keep them the same. And I'd always do both npm test and python test, even if there is no change in the given PR in the respective folder, since we want to have the ci.yml to be not just unit tests but integration tests as well

Issue #207, opened 2025-11-25
Expose config as a backend api endpoint


tmi:
A get endpoint for obtaining the frontend-relevant portion of the config, namely the auth (on/off) and if on then relevant details

liefra:
@tmi : As an initial draft for discussion of the config API spec, here is a proposal that, in my current view, offers a starting point. The aim is to include basic information about the environment, the backend API base URL for debugging, the language — although only English is planned for now — the auth type and an optional OIDC config (to be used when oidc auth type is active). I have also added a feature flag key-value pairs, which could be used to activate or deactivate specific features. It may not be needed at first, but it could offer useful scope for future extension. What do you think? Anything missing?

```ts

export const AUTH_TYPES = ['oidc', 'anonymous'] as const

export const ENVIRONMENTS = [
  'local',
  'development',
  'staging',
  'production',
  'test',
] as const


export const appConfigSchema = z.object({
  /** Environment name */
  env: z.enum(ENVIRONMENTS).default('local'),

  /** Backend API base URL */
  apiUrl: z.string().default('/api/v1'),

  /** Language code (ISO 639-1) */
  language: z.string().default('en'),

  /** Feature flags - extensible key-value pairs */
  featureFlags: z.record(z.string(), z.boolean()).default({}),

  /** Authentication type */
  authType: z.enum(AUTH_TYPES).default('anonymous'),

  /** OIDC configuration (required when authType is 'oidc') */
  oidcConfig: oidcConfigSchema.optional().nullable(),

  /** Debug mode flag */
  debug: z.boolean().default(false),
})

```

PR #211, opened 2025-11-27
[backend] add uiConfig endpoint

tmi:
https://github.com/ecmwf/forecast-in-a-box/issues/207


backend/forecastbox/api/routers/admin.py
```diff
@@ -184,3 +186,23 @@ async def patch_user(user_id: UUID4, update_dict: dict, admin: UserRead = Depend
         _ = await session.execute(query)
         await session.commit()
         return HTMLResponse(content="User updated successfully", status_code=200)
+
+
+@dataclass
+class ConfigResponse:
+    language_iso639_1: str
+    authType: Literal["anonymous", "authenticated"]
+    oidcConfig: OIDCSettings | None  # is None <=> authType is anonymous
```
liefra: Could be simplified to `loginEndpoint: str | None`

Issue #213, opened 2025-11-27
Prototype backend API for dynamic job building


tmi:
Basic endpoints as described in https://github.com/ecmwf/forecast-in-a-box/issues/208 with dummy data only

PR #214, opened 2025-11-27
[backend] prototype dynamic graph building

tmi:
Implements the API as proposed in https://github.com/ecmwf/forecast-in-a-box/issues/213, in particular:
- get catalog
- validate_expand
- compile

in a dummy manner, but in a way that frontend can integrate with

the actual implementation of the catalog and of some of the validation/expansion methods should follow more closely what is in https://github.com/ecmwf/forecast-in-a-box/tree/experimental/experimental. For now I've duplicated-trimmed a subset of what we will need, to allow integration on the frontend side to start -- but the frontend should not be affected by a complete change as long as we keep the contract in api/types/graph_building intact

### DIFF
--- a/backend/forecastbox/api/fable.py
+++ b/backend/forecastbox/api/fable.py
@@ -1,16 +1,22 @@
-from itertools import groupby
 from collections import defaultdict
-from forecastbox.api.types.graph_building import (
-    BlockKind,
-    BlockFactoryId,
-    BlockFactoryCatalogue,
-    GraphValidationExpansion,
-    BlockFactory,
-    BlockConfigurationOption,
-    GraphBuilder,
-)
-from forecastbox.api.types import RawCascadeJob
+from itertools import groupby
+
 from cascade.low.core import JobInstance
+
+from forecastbox.api.types import RawCascadeJob
+from forecastbox.api.types.fable import (
+    BlockConfigurationOption,
+    BlockFactory,
+    BlockFactoryCatalogue,
+    BlockFactoryId,
+    BlockKind,
+    FableBuilder,
+    FableValidationExpansion,
+)
+
+"""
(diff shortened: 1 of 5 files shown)


backend/forecastbox/api/graph_building.py
```diff
@@ -0,0 +1,145 @@
...
+            kind="source",
+            title="Compute Model Forecast",
+            description="Download initial conditions, run model forecast",
+            configuration_options={
+                "model": ActionConfigurationOption(title="Model Name", description="Locally available checkpoint to run", value_type="str"),
+                "date": ActionConfigurationOption(
+                    title="Initial Conditions DateTime", description="DateTime of the initial conditions", value_type="datetime"
+                ),
+                "lead_time": ActionConfigurationOption(title="Lead Time", description="Lead Time of the forecast", value_type="int"),
+                "ensemble_members": ActionConfigurationOption(
```
tmi: todo: allow for various configuration prefills, including both user prefills and admin mandates. It should be presumably doubly keyed, as in, given this ActionFactory with this value for this field, these fields are prefilled like this and these fields are not editable

liefra:
Just had another look at the graph_building API  as provided in this PR and do not have any immediate change requests for the API design. I think it should work, but will do more testing base on this API design before the next weekly meeting.

One question, though: the current flow is very much built around the `/graph_building/expand` endpoint, which can be used to configure a forecast job graph dynamically. That is fine, but I wonder how we handle different forecast-configuration modes, for example a quick-start mode with a recommended model, lead time and most frequently used products, or an option to select all available products automatically. To support this, we might need an additional endpoint that can provide predefined configuration graphs, or parts of graphs, to reduce the number of steps required. I guess this is more a question outside the scope of this PR; I just wanted to flag it for future discussion.

PR #225, opened 2025-12-17
[backend] upgrade ty

tmi:
Moving from alpha to beta, so worth the chore

A bunch of extra type: ignores, mostly on the sqlalchemy side

Two somehow-bugs caught & fixed

Three redundant casts removed


HCookie (approved):
I get the benefit of this tool, but I'm not a huge fan of the proliferation of type ignore's in the code base....

PR #229, opened 2025-12-31
[backend] fix datasetid bug

tmi:
for some dataset ids (too long, too slashy, proxy-mangled) we did not serve the results. It is a bad idea to have potentially arbitrary strings used in url paths

thus we move the dataset_id parameter from path param to request param

this is done by marking the original route deprecated, and introducing a new route for this. Additionally, we deprecate the route for result availability, that one is additionally potentially ambigous -- but its functionality is well served by another route

after frontend migrates, we delete the deprecated routes

### DIFF
--- a/backend/src/forecastbox/api/routers/job.py
+++ b/backend/src/forecastbox/api/routers/job.py
@@ -316,7 +316,7 @@ async def get_job_availability(job_id: JobId = Depends(validate_job_id), user: U
     Returns
     -------
     list[TaskId]
-        List of dataset IDs that are available for the job.
+        List of TaskIds that have an available output within this job.
     """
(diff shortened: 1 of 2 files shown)


backend/src/forecastbox/api/routers/job.py
```diff
@@ -425,7 +427,7 @@ def _build_zip() -> tuple[bytes, str]:
         )
 
 
-@router.get("/{job_id}/results/{dataset_id}")
+@router.get("/{job_id}/results")
```
tmi: cc @liefra -- this PR marks a route deprecated. If it gets merged, please at your convenience switch from dataset_id being a path-param to a query-param. I will then eventually delete the deprecated routes

HCookie (approved):

Issue #242, opened 2026-01-12
Plugin Installation: registry and backend APIs


tmi:
Implement the registry and backend APIs from https://github.com/ecmwf/forecast-in-a-box/issues/241

* [x] config of plugin store url
* [x] schema of plugin store
* [x] create the ecmwf plugin store file
* [x] retrieval of plugin store data
* [x] extend plugin metadata
* [x] plugin install based on store
* [ ] tool for creating plugin store entry (postponing for later, the scope doesnt justify it at the moment)

PR #251, opened 2026-01-14
[backend/plugins] installation and stores and metadata

tmi:
https://github.com/ecmwf/forecast-in-a-box/issues/242

It's not fully tested, because we don't have any plugins actually published (in particular to pypi). But at least some API is in place, and I tested locally some basics, with data mocked

### DIFF
--- a/backend/packages/fiab-core/src/fiab_core/fable.py
+++ b/backend/packages/fiab-core/src/fiab_core/fable.py
@@ -70,6 +70,12 @@ class PluginCompositeId(BaseModel):
 
 
 class PluginBlockFactoryId(BaseModel):
+    """Note to plugin authors: This is a routing class. When you implement your BlockFactories for the catalogue,
+    you dont use this, you only need to declare a BlockFactoryId unique inside your plugin. Similarly, when you
+    return which BlockFactories are possible in the expand method, you only return your BlockFactoryIds. This
+    appears only when you receive BlockInstances in the compile/validate -- and again, you just need to use the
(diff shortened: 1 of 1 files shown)


backend/packages/fiab-core/src/fiab_core/fable.py
```diff
@@ -48,10 +49,28 @@ class BlockFactory(BaseModel):
...
+        return cls(store=store, local=local)
+
+    @staticmethod
+    def to_str(k: Self) -> str:
+        return f"{k.store}:{k.local}"
 
 
 class PluginBlockFactoryId(BaseModel):
-    plugin: PluginId
+    plugin: PluginCompositeId
```
tmi: @jinmannwong this is the change I mentioned yesterday that may cause some conflicts with your PR. But thinking of it now, I don't like it, I don't want plugins to declare what store they will be published in... I'll need to change it so that inside plugins only PluginId is used :/ But the rest of PR should be fine already

tmi: hah, I actually confused myself. Upon deeper inspection, I realized that even though PluginStoreId appears here, the plugin author does not use it when implementing the plugin. It's not immediately obvious, so I added this (hopefully) clarifying doc https://github.com/ecmwf/forecast-in-a-box/pull/251/changes/adb52e1dfcc8cb05e759f526bb5846e2af55e795 . That should also mean that the possible PR conflict should be smaller than I thought

PR #255, opened 2026-01-21
[backend/plugins] Initial ECMWF plugin

corentincarton:
### Description

Small architecture proposal for plugin (not tested)

### Contributor Declaration

### DIFF
--- a/backend/pyproject.toml
+++ b/backend/pyproject.toml
@@ -81,7 +81,7 @@ dev = [
   "fiab-plugin-toy", # TODO remove after first publish
   "fiab-plugin-toy-impl", # TODO remove after first publish
   "fiab-plugin-toy2[runtime]", # TODO remove after first publish
-  "fiab-plugin-ecmwf", # TODO remove after first publish
+  "fiab-plugin-ecmwf[runtime]", # TODO remove after first publish
 ]
 
 [tool.setuptools_scm]
--- a/backend/uv.lock
+++ b/backend/uv.lock
@@ -1772,7 +1772,7 @@ webmars = [
 
 [package.dev-dependencies]
 dev = [
-    { name = "fiab-plugin-ecmwf" },
+    { name = "fiab-plugin-ecmwf", extra = ["runtime"] },
     { name = "fiab-plugin-toy" },
     { name = "fiab-plugin-toy-impl" },
     { name = "fiab-plugin-toy2", extra = ["runtime"] },
@@ -1825,7 +1825,7 @@ provides-extras = ["all", "plots", "thermo", "webmars"]
 
 [package.metadata.requires-dev]
 dev = [
-    { name = "fiab-plugin-ecmwf", editable = "packages/fiab-plugin-ecmwf" },
+    { name = "fiab-plugin-ecmwf", extras = ["runtime"], editable = "packages/fiab-plugin-ecmwf" },
     { name = "fiab-plugin-toy", editable = "packages/fiab-plugin-toy" },
     { name = "fiab-plugin-toy-impl", editable = "packages/fiab-plugin-toy-impl" },
     { name = "fiab-plugin-toy2", extras = ["runtime"], editable = "packages/fiab-plugin-toy2" },


backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
```diff
@@ -0,0 +1,223 @@
...
+        )
+
+
+ensembleStatistics = EnsembleStatisticsFactory(
+    title="Ensemble Statistics",
+    description="Computes ensemble mean or standard deviation",
+    configuration_options={
+        "variable": BlockConfigurationOption(
+            title="Variable", description="Variable name like '2t'", value_type="str"
+        ),
```
HCookie: We need to be able to limit this to the variables of the source, what are the thoughts on that?

backend/packages/fiab-core/src/fiab_core/plugin.py
```diff
@@ -8,41 +8,58 @@
...
-
-Compiler = Callable[
-    [DataPartitionLookup, BlockInstanceId, BlockInstance], Either[DataPartitionLookup, Error]  # type:ignore[invalid-argument] # semigroup
-]
-"""Given a cascade builder and a block instance corresponding to this plugin's Factory, either update the builder with corresponding tasks or provide error"""
-# NOTE JobBuilder + DataPartitionLookup to be replaced with Fluent
 
 
 @dataclass
 class Plugin:
```
corentincarton: Main architectural change of this PR, but I think it plays nicely with the plugin. We keep the flexibility to change those functions by overriding the class, but I don't think it should happen.

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
```diff
@@ -0,0 +1,223 @@
...
+    },
+    inputs=["dataset"],
+)
+
+
+def write_zarr(dataset, path: str) -> None:
+    dataset.to_zarr(path, mode="w")
+
+
+class ZarrSinkFactory(SinkFactory):
```
corentincarton: @jinmannwong, try to stay compatible with this if you add more blocks. It would be nice to have an end-to-end pipeline to test.

jinmannwong: Should I work on your branch? If I add more blocks on the other branch then we would have to be re-implemented them in the way you do here.

backend/packages/fiab-core/src/fiab_core/fable.py
```diff
@@ -104,3 +104,55 @@ class XarrayOutput(BaseModel):  # NOTE eventually Qubed
...
+    kind: Literal["transform"] = "transform"
+
+    def validate(self, block: "BlockInstance", inputs: dict[str, "BlockInstanceOutput"]):
+        raise NotImplementedError
+
+    def compile(self, partitions: "DataPartitionLookup", block_id: "BlockInstanceId", block: "BlockInstance"):
+        raise NotImplementedError
+
+    def intersect(self, output: "BlockInstanceOutput") -> bool:
+        return False
```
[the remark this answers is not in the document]
jinmannwong: If we go down this route, can these methods not be moved into the definition of `BlockFactory` itself? I think actually they should all raise the `NotImplementedError` except possibly for `SourceFactory.intersect` which should return `False`.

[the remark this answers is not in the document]
corentincarton: Yes, BlockFactory should provide the interface but return NotImplementedError.

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/runtime/__init__.py
```diff
@@ -0,0 +1,4 @@
+import earthkit.data
+from earthkit.workflows.backends import register
+
+register(earthkit.data.SimpleFieldList, earthkit.workflows.backends.earthkit.FieldListBackend)
```
tmi: one note of caution: this imposes the constraint that every function which is an entrypoint must import this file, otherwise execution of this line is not guaranteed, because various tasks can be executed each in their own process. But processes may be re-used, so this possibly affects other tasks. But I don't think registering a backend is a dangerous operation, esp since the registered implementation is guaranteed to be importible in this case

install/plugins.json
```diff
@@ -9,7 +9,7 @@
             "display_description": "Example Fable Workflows",
             "display_author": "ECMWF"
         },
-        "ecmwf": {
+        "base": {
```
tmi: I'd probably stick to ecmwf-base... its an internal id so wont be user visible unless there is a keyerror or something, but still

PR #263, opened 2026-02-03
[cd] improve installer

tmi:
- dont always force update, only when envvar or file
- when updating via api, carry out the launcher etc update -- on the next launch
- detect that first-time install happened
- upon first-time install, install plugin
- dont redirect to catalog upon plugin changes -- to prevent a 530
- improve testing a release candidate locally

with this branch merged, the fiab.sh launcher should be working quite neatly. It additionally supports updating itself -- so if things go well, this is the last time the users will need to replace fiab.sh by hand

### DIFF
--- a/backend/src/forecastbox/api/updates.py
+++ b/backend/src/forecastbox/api/updates.py
@@ -112,4 +112,5 @@ def mark_release(release: Release) -> None:
 
 
 def should_install_default_plugin() -> bool:
-    return Globals.FirstRun
+    return False  # TODO enable post-LTS release
+    # return Globals.FirstRun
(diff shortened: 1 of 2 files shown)


backend/src/forecastbox/api/routers/plugin.py
```diff
@@ -96,8 +96,10 @@ def get_plugin_details(forceRefresh: bool = False) -> PluginListing:
     return PluginListing(plugins=rv)
 
 
-# TODO get_catalogue, or fable:get_catalogue, or what?
-get_catalogue_redirect = lambda request: RedirectResponse(request.url_for("get_catalogue"), status_code=status.HTTP_303_SEE_OTHER)
+# TODO ideally we'd return the redirect here, but that is basically guaranteed to end up with a 503 because
+# the plugins aren't ready yet -- we probably need to await here or smth
+# get_catalogue_redirect = lambda request: RedirectResponse(request.url_for("get_catalogue"), status_code=status.HTTP_303_SEE_OTHER)
+get_catalogue_redirect = lambda request: Response(status_code=200)
```
liefra: Maybe you could respond with a `202` (Accepted) first to be more precise that the request was not processed yet (and actually it is not yet clear if the install with succeed)

PR #267, opened 2026-02-05
[fable] feat!: Add QubedInstanceOutput for ECMWF Plugin

HCookie:
### Description
Expand BlockInstanceOutput with Qubed.

Use `qubed` as the dataqube layer with other ways to set broad keys

```python
output = QubedInstanceOutput(dataqube=Qube.from_datacube({
     'param': ['2t', 'tp'],
     'time': [0, 1, 2],
     'level': [1000, 850, 700],
}))
output.dimensions()
#{'param', 'time', 'level'}
expanded = output.expand({'ensemble': ['ens1', 'ens2']})
expanded.dimensions()
#{'param', 'time', 'level', 'ensemble'}
```

### Contributor Declaration


backend/packages/fiab-core/src/fiab_core/tools/plugins.py
```diff
@@ -23,8 +23,6 @@ def validate(self, block: BlockInstance, inputs: dict[str, BlockInstanceOutput])
 
     def expand(self, block: BlockInstanceOutput) -> list[BlockFactoryId]:
         """Given a block instance output (including from other plugin), provide which block factories from this plugin can expand it"""
-        if len(block.variables) == 0:
-            return []
```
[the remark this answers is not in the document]
tmi: additionally, fiab-core needs to come from fiab itself, so whenever this happens, we need full fiab release

backend/packages/fiab-core/src/fiab_core/fable.py
```diff
@@ -95,12 +96,268 @@ class BlockInstance(BaseModel):
...
+
+    model_config = ConfigDict(frozen=True, extra="allow")
+
+    datatype: str = Field(default="")
+    """e.g. 'xarray', 'grib', 'netcdf', 'plot'"""
+
+    # TODO: Add more fields as needed.
+
+
+class BlockInstanceOutput(BaseModel):
```
tmi: firstly, this hardcodes every block instance output to be qubed, we dont want that, because of image outputs and the like.

secondly, while currently BlockInstanceOutput is a short-lived class only, ie, it is not being serded to frontend, I'm still worried about making it contain an object like this. There is a high chance we would be caching it, passing across threads, etc. Is it safe, will it be safe?

thirdly, I don't think this bunch of forwarder-wrapper methods do nothing but bloat this file with docstrings.

so what I think would be best is:
```
BlockInstanceOutput = Qubed | Any
```

and in the expand step, you would match on the block instance output type

and all the expand, collapse, contains, whatever methods, would go to Qubed itself, I don't think they are legitimate on BlockInstanceOutput level

backend/packages/fiab-core/pyproject.toml
```diff
@@ -19,10 +19,13 @@ dynamic = [ "version" ]
 dependencies = [
     "pydantic",
     "earthkit-workflows", # NOTE technically we would need just cascade.low.core and some fluent.core, not the whole runtime -- possibly split down there as well
+    "qubed>=0.3.0",
 ]
 
 [dependency-groups]
 dev = [
+  "pytest>=8.4.2",
+  "pytest-xdist>=3.8.0",
```
tmi: this is odd, why?

backend/packages/fiab-plugin-toy/src/fiab_plugin_toy/__init__.py
```diff
@@ -21,7 +21,6 @@
     BlockInstanceId,
```
tmi: btw dont worry about toy plugins, feel free to drop them in this PR

backend/packages/fiab-core/src/fiab_core/fable.py
tmi: > I do not think these methods can be moved to qubed itself.

then we can have a Qubed utility wrapper or something. But not on the BlockInstanceOutput member -- because why would we unify the concern of "passing schema from one block to another" with the concern of "operating a qubed object"?

HCookie:
. . .

backend/packages/fiab-core/src/fiab_core/fable.py
```diff
@@ -95,19 +95,43 @@ class BlockInstance(BaseModel):
     """Keys come from factory's `inputs`, values are other blocks in the (partial) fable"""
 
 
-class XarrayOutput(BaseModel):  # NOTE eventually Qubed
+class BlockInstanceOutput(Protocol):
+    def is_empty(self) -> bool: ...
+    def dimensions(self) -> set[str]: ...
```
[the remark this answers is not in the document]
HCookie: The complex data cubes I envision flowing through this system cannot be expressed simply with the `XarrayOutput`, the qubed object maps more directly to the datatree used in fluent, and the data possible with `ekd.Fieldlist's`. I allowed for a dict usage to simplify usage for other authors, as I value that part of my code. Let's discuss at some point in the future, what a proper minimal could be, and maybe it is add qubed to the base object, or do the `fields`, with `hasattr`.

PR #274, opened 2026-02-12
Refactor fable plugin test

jinmannwong:
### Description
Change nonpytest to use fiab-plugin-ecmwf and add smaller integration test that reads from file. Needs new release of fiab-plugin-ecmwf that contains small improvement to retrieval function.

### Contributor Declaration

### DIFF
--- a/backend/tests/integration/test_fable.py
+++ b/backend/tests/integration/test_fable.py
@@ -32,7 +32,7 @@ def test_fable_contruction(tmpdir, backend_client_with_auth):
     blocks = {"source1": source}
     builder = FableBuilderV1(blocks=blocks)
     response = backend_client_with_auth.request(url="/fable/expand", method="put", json=builder.model_dump())
-    assert len(response.json()["possible_expansions"]["source1"]) == 3
+    assert len(response.json()["possible_expansions"]["source1"]) > 0
 
     temporalMean = BlockInstance(
         factory_id=PluginBlockFactoryId(plugin=pluginId, factory="temporalStatistics"),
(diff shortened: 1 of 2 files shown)


backend/src/forecastbox/config.py
```diff
@@ -171,7 +171,7 @@ class ProductSettings(BaseModel):
     default_input_source: str = "opendata"
     """Default input source for models, if not specified otherwise"""
 
-    plugins: PluginsSettings = Field(default_factory=dict)
+    plugins: PluginsSettings = Field(default_factory=_default_plugins)
```
jinmannwong: @tmi I'm aware I should change this. Would the correct way to handle this be to call `install_default_plugins`?

jinmannwong: > @jinmannwong I made one bugfix (the json of job/outputs does not have output_ids field, it is the output_ids itself already) and tweaked the sleep/attempt counts (the 2 minute wait interval was too long, it meant the test would kept running up to 40 minutes if something went wrong). Now its passing locally for me e2e

Thanks! Did you not encounter the job status being "failed" due to a `TimeoutError` then?

jinmannwong: > but isnt there something smaller? Downloading 35MB as a part of unit test seems... excessive. Thats why I originally had it in nonpytest -- maybe we could keep it as nonpytest, and have in pytest only something smaller and faster? Possibly with mocked data, like we do for the model download test?

Sorry, yes, I agree. I wanted to make sure everything was running correctly first. I will change it so we fetch the data from file.

jinmannwong: Ah, ok, thanks.

jinmannwong: > what do you mean by information about data contents -- like schema, column names, partition count? 

Yes, if I understand you correctly. I meant the different dimensions in the data, e.g. number of parameters, steps, is it ensemble or not. 

> And why is it popping up in the context of the plugin test? Aren't we fully in control of what file we provide to the test?

Maybe we were not thinking of the same thing regarding using a file for the test as input. I assumed this meant creating `FileSource` fable block that can handle user provided files. Were you thinking that we mock the ecmwf-open-data retrieval by reading from a file instead?

tmi: (for the record) yy, mock open data retrieval by reading a local file

backend/tests/integration/utils.py
```diff
@@ -38,3 +41,21 @@ def prepare_cookie_with_auth_token(token) -> dict:
         A dictionary representing the cookie with the token.
     """
     return {"name": "forecastbox_auth", "value": token}
+
+
+def ensure_completed(backend_client, job_id, sleep=0.5, attempts=20):
+    i = attempts
```
tmi: out of curiosity, why do we need to introduce the new variable `i` here?

PR #279, opened 2026-02-18
[backend] update dataclass usage

tmi:
Just adding frozen and slots where it makes sense, and updating AGENTS.md to stick to it too


tmi:
cc @HCookie -- I'm merging right away as its trivial, but feel free to add reviewer comments and I'll fix in a follow up

PR #295, opened 2026-02-25
Feat/runtime model deps

tmi:
Extends the EnvironmentSpecification with list of CompositeCheckpointId, downloads them (naively!) prior to job launch


backend/src/forecastbox/api/artifacts/base.py
```diff
@@ -14,17 +14,7 @@
 from dataclasses import dataclass
 from pathlib import Path
 
-from fiab_core.artifacts import MlModelCheckpoint, MlModelCheckpointId, Platform
-
-from forecastbox.config import ArtifactStoreId
```
tmi: now there is duplicate ArtifactStoreId -- once in fiab_core.artifacts, once in forecastbox.config -- remove the latter one and replace it with import from fiab_core

PR #300, opened 2026-03-06
Chore/remove old endpoints

tmi:
Removes all model/ and product/ routes

Removes model db

Refactors execution/execute route into jobs/ router, and removes the remaining execution/ routes. The corresponding change in frontend also done

This should capture the majority of v1 deprecated code being removed. There remains a ProductToOutputId which we use to extract outputs ids from any job -- so it cant be deprecated right away, though the name and 2 fields are unfit. I think we'll be making further changes to this later on, so for now leaving without change

@HCookie -- if there was anything worth preserving, now is the time to speak :)

@liefra -- I used api/endpoints.ts as the authoritative source of what you are using (btw much appreciate that it exists as a single file, that is very neat to work with!), pls correct me if that was not the case. And if I should have done something more than just the "execution" -> "job" string replacement :)


liefra:
> I used api/endpoints.ts as the authoritative source

@tmi : You found the correct file :) `api/endpoints.ts` file is the authoritative single source containing all backend APIs currently in use and referenced by other files. Currently no model or product backend routes are currently in use.

PR #355, opened 2026-04-02
Add variable interpolation for workflow configuration values

tmi:
Step 1 of variable support -- automatic variables + interpolation

In detail

Introduces ${variable} interpolation in BlockInstance configuration values, with a set of automatic (system-provided) variables available at run time.

New modules:
- domain/variables/automatic.py: AvailableAutomaticVariables literal and get_values_and_examples() for pre-submit validation / frontend display
- domain/variables/resolution.py: extract_variables() and resolve_configurations() for parsing and substituting ${var} patterns

Changes:
- routes/blueprint.py: VariableDetail model and GET /variables/list endpoint with assert_never-based exhaustive match over AvailableAutomaticVariables
- domain/blueprint/service.py: validate_expand() now resolves known variables before invoking the plugin validator, and errors on unknown variable references
- domain/run/compile.py: resolve_automatic_values() maps runId/submitDatetime to runtime values; compile_builder() accepts variable_values and resolves them before compiling each block
- domain/run/service.py: execute() pre-generates run_id so it can be used as a variable value and passed to both compile_builder() and upsert_run()
- domain/run/db.py: upsert_run() now accepts a pre-generated run_id for new runs (attempt 1 is created when the id does not exist yet)

Tests:
- unit/test_variables_resolution.py: unit tests for extract_variables and resolve_configurations (success and edge cases)
- integration/test_blueprint.py: test_list_available_variables, updated test_blueprint_expand to verify unknown-variable rejection and known-variable acceptance, updated test_blueprint_basic_execute to use ${runId} in fname
- unit/test_jobs.py: updated test to reflect new upsert_run semantics

### Contributor Declaration


backend/src/forecastbox/domain/blueprint/service.py
```diff
@@ -129,6 +131,14 @@ def validate_expand(blueprint: BlueprintBuilder) -> BlueprintValidationExpansion
             # TODO most likely disable this, we would inject defaults at the compile level
             block_errors[blockId] += [f"Block contains missing config: {missingConfig}"]
 
+        extracted_vars = variable_resolution.extract_variables(blockInstance).t
```
tmi: you should check out for `e` being None first -- if it is not None, then append it to block_errors and continue, otherwise carry on

backend/src/forecastbox/domain/blueprint/service.py
```diff
@@ -129,6 +131,14 @@ def validate_expand(blueprint: BlueprintBuilder) -> BlueprintValidationExpansion
...
+        available_vars = set(get_values_and_examples().keys())
```
tmi: don't evaluate this anew at the start of every iteration step -- do it once, prior to entering for cycle

backend/src/forecastbox/domain/blueprint/service.py
```diff
@@ -129,6 +131,14 @@ def validate_expand(blueprint: BlueprintBuilder) -> BlueprintValidationExpansion
...
+        unknown_vars = cast(set[str], extracted_vars) - available_vars
```
tmi: make the cast at the callsite where `extracted_vars` is defined

backend/src/forecastbox/domain/run/compile.py
```diff
@@ -35,20 +53,24 @@ def _get_artifacts_list(graph: Graph) -> list[CompositeArtifactId]:
     return list(artifacts)
 
 
-def compile_builder(blueprint: BlueprintBuilder) -> ExecutionSpecification:
+def compile_builder(blueprint: BlueprintBuilder, variable_values: dict[str, str] | None = None) -> ExecutionSpecification:
```
tmi: there should be no need to allow for None in the type -- we should always provide a value at the callsite, possibly {} but thats ok

backend/src/forecastbox/domain/variables/__init__.py
```diff
@@ -0,0 +1,8 @@
+# (C) Copyright 2024- ECMWF.
```
tmi: empty files dont need licenses

backend/src/forecastbox/domain/variables/automatic.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def get_values_and_examples() -> dict[AvailableAutomaticVariables, str]:
+    """Return all automatic variable names with hardcoded example values.
+
+    Used for pre-submit validation and frontend display. The examples are stable
+    representative values, not generated at runtime.
+    """
+    return {
+        "runId": "run-abc123-def456",
```
tmi: this does not look like uuid.uuid4() str invocation, which is what is being used for run id. You best do this as follows -- declare a top level variable values_and_examples, with keys like here but values being generated at import time, that is `str(uuid.uuid4())` and strftime of datetime now. And then this function just returns that variable. Add a `# TODO` there with "replace with frozendict once available so that we have immutability"

backend/tests/unit/domain/variables/test_resolution.py
```diff
@@ -0,0 +1,118 @@
+# (C) Copyright 2024- ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Unit tests for domain/variables/resolution."""
```
tmi, line 10: organize this file accordingly, that is unit/domain/variables/test_resolution.py

backend/src/forecastbox/routes/blueprint.py
```diff
@@ -282,3 +290,19 @@ def expand_blueprint(blueprint: BlueprintBuilder) -> BlueprintValidationExpansio
         possible_sources=result.possible_sources,
         possible_expansions=result.possible_expansions,
     )
+
+
+@router.get("/variables/list")
+def list_available_variables() -> list[VariableDetail]:
```
tmi, line 296: cc @liefra so we are starting for real with variables. This is a new endpoint which will show which automatic (or maybe "implicit", or "system" would be a better name) variables are available in general, its a sort of catalogue. Maybe we could use it in some help or doc or autocomplete or something later. Feel free to suggest any changes or improvements to this. I can imagine adding user-defined and system-level custom variables here later once we have a support for them on the backend, maybe distinguish them via some `source` enum (both as optional filter and as a return type discriminator)

for now only minimum is supported -- we have two of those variables, and we only support the basic `${variableName}` substitution -- no eval of actual python code, function invocations or anything. This will change later ofc. I've added integration tests which check that the `${runId}` indeed evaluates and is used by the actual job execution

I don't think this creates any immediate follow ups for you, just a heads up

PR #358, opened 2026-04-08
Remove dynamic_expr from frontend

tmi:
@liefra could you please peek at this one? I've removed the `dynamic_expr` field from the Schedule entity, as we will now be using variables solely. Ie, you don't put the dynamic expression in the UI, you simply use `${submitDatetime}` in the respective configuration options

I'm a bit puzzled that this PR deletes so much things -- based on a brief skim, is it legit, can I just merge it?

(Heres' the bigger PR which does the removal itself on the backend side, and the required work on the variables themselves, https://github.com/ecmwf/forecast-in-a-box/pull/356, but there are no frontend changes there)


liefra:
@tmi : I have reviewed the changes, they are all valid. You can merge them.

PR #365, opened 2026-04-10
Poc/interpolation

tmi:
Implements the Jinja interpolation with the ${expr} syntax, and a bunch of UDFs, as documented by a newly added endpoint `blueprint/glyphs/functions`. Backwards compatible with all expressions that were valid before.

cc @liefra -- lmk if the new endpoint is fine enough. Given that the list of glyphs is inherently dynamic, your part must actually be dynamic, so yeah, going with just function list via an endpoint. I don't expect we'll have _grammar_ changes going forward, so maybe this will be fine for a while.


backend/src/forecastbox/domain/glyphs/jinja_interpolation.py
```diff
@@ -0,0 +1,137 @@
+# (C) Copyright 2024- ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Jinja2-based string interpolation engine for ${...} glyph expressions.
```
tmi, line 10: this file should be named `jinja_interpolation`

backend/src/forecastbox/domain/glyphs/_interpolation.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def _collect_glyph_names(node: jnodes.Node, glyphs: set[str]) -> None:
+    if isinstance(node, jnodes.Name) and node.ctx == "load" and node.name not in _FILTER_NAMES:
+        glyphs.add(node.name)
+    for child in node.iter_child_nodes():
+        _collect_glyph_names(child, glyphs)
+
+
+def extract_glyph_names(raw: str) -> tuple[set[str], str | None]:
```
tmi: use the class Either for the return type -- look for its usage across the codebase

PR #375, opened 2026-04-16
Feat/improve plugin builders

tmi:
1/ Moves the qube + metadata to fiab-core
2/ Moves the qube + metadata utility methods to a dedicated submodule in fiab-plugin-ecmwf
3/ Updates the plugin tools to work with QubedOutput only, to save a ton of isinstance checks
4/ Promotes the is_empty optimization to fiab backend itself

I did not eliminate all `# type: ignores` that have now become redundant -- but most has, now it should be actually type-wise correct

I think we should have had the `tools` submodule in fiab-plugin-ecmwf instead of fiab-core as well, but fine with sticking to that for now


backend/packages/fiab-core/src/fiab_core/tools/blocks.py
```diff
@@ -20,18 +20,19 @@
     BlockInstanceId,
     BlockInstanceOutput,
     BlockKind,
+    QubedOutput,
 )
 from fiab_core.plugin import Error
 
 
-class BlockBuilder(abc.ABC):
+class QubedBlockBuilder(abc.ABC):
```
corentincarton, line 28: I would keep this as BlockBuilder, no need to be too verbose and put Qubed everywhere.

tmi: (for the record: BlockBuilder won't appear at many places, and the actually-used children of it remain as just Source, Transform, Product, Sink. And for the other case, QubedPluginBuilder, we'll have one ocurrence per plugin so its also kinda ok)

tmi:
@HCookie this PR would trigger some reworks in the Anemoi-Plugin PR. I'm happy to do them, once this PR is reviewed & merged. But if you want to do them yourself, I'm happy with that as well :)

PR #381, opened 2026-04-16
feat(glyphs): support glyph values that reference other glyphs

tmi:
@corentincarton / @liefra -- we now support glyphs within glyphs, that is, you can declare $dataRoot=$globalRoot/$personalSuffix/$runId, etc, and then use $dataRoot in the configuration values in builder

it is moderately tricky how this plays out with validations and with restarts/persistence, so I cant rule out there arent some subtle bugs. But in this PR I increased test coverage, and nothing existing broke, so basics should be working fine

# Details:

Add expand_glyph_values() to resolution.py which uses DFS to expand ${...} patterns inside glyph values before they are applied to block configuration options.  This enables composite glyphs such as:

    local_glyphs = {"myPath": "${root}/${runId}"}

where 'root' is a global glyph and 'runId' is intrinsic.

Key design decisions:
- DFS expansion detects self-references and mutual cycles and raises GlyphCircularReferenceError instead of silently leaving unresolved placeholders.
- Unknown references (keys absent from the glyph map) are kept as-is so normal block-level unknown-glyph validation still fires.
- background.py persists pre-expansion glyph values in the run context so composite glyphs re-expand correctly on restart, allowing pinned intrinsics (startDatetime, attemptCount) to refresh as expected.
- blueprint/service.py (validate_expand) catches circular reference errors and surfaces them as global_errors so the UI can report them.

Tests added:
- Unit: expand_glyph_values covering plain values, single/multi-level chains, unknown refs, self-cycles, mutual cycles, longer cycles.
- Integration: validate_expand reflects composite glyph expansion in resolved_configuration_options and reports circular refs as global_errors; execute test verifies runtime output is correctly expanded.

### DIFF
--- a/backend/src/forecastbox/domain/blueprint/service.py
+++ b/backend/src/forecastbox/domain/blueprint/service.py
@@ -139,8 +139,8 @@ async def validate_expand(
     global_glyphs = {str(row.key): str(row.value) for row in await global_db.list_global_glyphs(auth_context)}
     local_glyphs = blueprint.local_glyphs
 
-    all_glyphs = merge_glyph_values(intrinsic_values, global_glyphs, local_glyphs, {})
-    available_glyphs = set(all_glyphs.keys())
+    all_glyphs_raw = merge_glyph_values(intrinsic_values, global_glyphs, local_glyphs, {})
+    available_glyphs = set(all_glyphs_raw.keys())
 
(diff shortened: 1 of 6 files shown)


liefra:
Very cool. Will give it a try! Thanks

PR #387, opened 2026-04-21
Feat/user glyphs

tmi:
Moves form glyphs being uniquely constrained by `name` to `(name, created_by)`, and additionally adds `overriddable` setting for public glyphs.

This has the following UX effects:
1/ admin defines a public overiddable glyph like "default_model_checkpoint" for ease of use, but users can change their own personal default model
2/ admin defines a public nonoverriddable glyph like "maximum_memory" for environment safety, users cannot change this value. (But technically, when admin changes this value and user re-executes a past workflow, the original value is intact -- we prefer this for consistency/soundness reasons)

(cc @corentincarton -- lmk if changes required... I just wanted to get rid of the scenario where users "block" some glyph name for everybody in the original implementation, I dont think we should overcomplicate this)

There are some *Client Contract Changes*:
- the new flag overriddable is exposed, but can only be set when public=True, ie, by admins
- listing of glyphs now returns created_by, and lists *all* visible glyphs to the user. The users does not see the resolution priority, but that is best done when building a workflow anyway (where the final resolved value is shown)

(cc @liefra -- lmk if you want any changes here)


liefra:
@tmi : frontend-wise no changes are required, will then make use of the added `overriddable`

PR #399, opened 2026-04-23
test: add integration test for jinja interpolation via /expand endpoint

tmi:
Add test_blueprint_jinja_interpolation_expand to verify:
- A malformed jinja expression (invalid filter name with hyphens) causes block_errors (Jinja2 parses hyphens as subtraction, treating is, no, filter as unknown glyph variables)
- A well-formed pure-jinja expression using registered globals/filters (datetime, timedelta, floor_day) resolves to the correct value

Fix two related bugs found during implementation:
- extract_glyphs: pure-jinja expressions with only globals/filters and no glyph variable names were not added to glyphed_options, so their rendered values never appeared in resolved_configuration_options. Fixed by checking for any dollar-brace opener in the value when result.t is empty.
- validate_expand: runtime errors from render_expression (e.g. AttributeError from operator precedence issues) were unhandled, causing 500 responses. Wrapped resolve_configurations in try/except and surface as block_errors.

### DIFF
--- a/backend/src/forecastbox/domain/blueprint/service.py
+++ b/backend/src/forecastbox/domain/blueprint/service.py
@@ -198,7 +198,12 @@ async def validate_expand(
             block_errors[blockId] += [f"Unknown glyphs referenced: {unknown_glyphs}"]
             invalidable.add(blockId)
             continue
-        resolution.resolve_configurations(blockInstance, all_glyphs)
+        try:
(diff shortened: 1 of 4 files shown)


tmi:
cc @liefra -- this fixes the bug you reported with malformed jinja to /expand going with 500 -- it now gives 200 but the errors are included in the block error messages. And another bug was discovered -- the endpoint now resolves interpolation that contained only base jinja but no glyphs

PR #402, opened 2026-04-24
[fiab-plugin-ecmwf] fix: pass input_source to anemoi env builder

liefra:
### Description

- `AnemoiBuilder.build()` does not pass the user-selected input_source through. As a result, forecasts using the `opendata` source failed at runtime.

- Add `input_source: str | None = None` kwarg to `AnemoiBuilder.build` and forward it to `get_environment`.
- `AnemoiSource.compile` now passes `configuration["input_source"]` when building the Inference.

### Contributor Declaration

By opening this pull request, I affirm the following:

* All authors agree to the [Contributor License Agreement](https://github.com/ecmwf/codex/blob/main/Legal/Contributor-License-Agreement.md).
* The code follows the project's coding standards.
* I have performed self-review and added comments where needed.
* I have added or updated tests to verify that my changes are effective and functional.
* I have run all existing tests and confirmed they pass.

PR #405, opened 2026-04-24
fix: Allow for full split from anemoi-inference

HCookie:
### Description
Store qubed in artifacts rather than large metadata
- Elevate environment retrieval to top (alternate approach from #402)
- Remove unneeded `pproc` deps

### Contributor Declaration


HCookie:
@tmi The use of the MLCheckpoint went further then I thought, what do you think of the other uses and progating the new attributes?

.github/workflows/ci.yml
```diff
@@ -88,7 +88,7 @@ jobs:
           set -euo pipefail
           pushd backend
           export UV_FROZEN=1
-          uv sync --extra plots --python "${{ matrix.python_version }}" --all-packages
+          uv sync --extra runtime --python "${{ matrix.python_version }}" --all-packages
```
tmi, line 91: ok with this, but there is a better way -- separate the pytests of ecmwf plugin into say runtime and unit, and do the `just val` in this package as `uv run pytest ./tests/unit && uv run --with .[runtime] pytest ./tests/runtime`. That should be more robust toward any sort of leakage issues

if you are ok with maintaining the tests separation into unit and runtime long term, I'd be happy to set it up like this

tmi:
> @tmi The use of the MLCheckpoint went further then I thought, what do you think of the other uses and progating the new attributes?

well on the backend we simply pass it up to the frontend, which displays that on the ArtifactsDetail page. The field was meant to be provisionary, see the (in-this-PR-deleted) NOTE above it. We have three ways of proceeding:
 - change the frontend in this PR to the desired state, display what we long term want for ArtifactDetails
 - change the frontend by simply replacing output_characteristics with output_qube and hope it will display well
 - hotfix by setting `output_characteristics = [str(output_qube)]`, ie, the frontend contract won't change, just the content

I'd go with the third for this PR and first long term (with a question who is responsible for building the displayable data -- it can be either the backend, or it can be the registry itself, ie, the displayable data would already be in the json in a dedicated category)

.github/workflows/ci.yml
HCookie: Sure, I think that could work

HCookie: @tmi Can you confirm I did it correctly?

HCookie:
tests failing due to old rawgithubcontent asset. (I think)

tmi (approved):

PR #406, opened 2026-04-27
[spec] add design proposal for outputs

tmi:
Proposing how to go forward with better output handling

### DIFF
--- a/docs/developer/changeSpecs/backend-mime_and_downstream.md
+++ b/docs/developer/changeSpecs/backend-mime_and_downstream.md
@@ -1,0 +1,32 @@
+# Goal
+Propose output type / mime type contract offered by the backend.
+Propose how integration with downstream postprocessors and tools like SkinnyWMS would be utilized.
+
+# MIME type
+A block instance can currently output multiple types: qubed, raw, none; possibly expanding in the future.
+The raw type can hold images like png, or documents like pdf.
+To determine the underlying type, we currently need to deserialize the raw bytes at the backend, and look up based on class or type what the correct mime type should be.
+
+**Proposal**: extend the Raw type with optional MIME type string field, and set it at the fable compile time.
+1. *Assumption*: we don't expect qubed or future possible types to be ever interpretable by the browser, or if they are, their MIME type to be static.
+2. *Assumption*: we expect MIME type to be determinable at the compile time, ie, without access to the runtime.
+3. *Assumption*: Raw type would have a single field, but the underlying fluent action / cascade subgraph can produce multiple outputs (either via a generator cascade node, or by being a static subgraph). All these outputs would be of the same MIME type.
+
+# SkinnyWMS
+For start, we can create a new domain/route set, like Downstream or PostprocessingTools or DisplayTools or Lenses.
+SkinnyWMS would be a first such, with routes like `start(location) -> handle(pid, port)`, `list()`, `kill(pid)`, `status(pid)`, etc.
+However, the question is how to go from "a job produced data at folder F" to "skinnyWMS has been launched at folder F":
+1. *Option*: the user does it manually -- the job outputs the location as RawOutput with mime type text, the user copies it and invokes the standalone `start` api call manually.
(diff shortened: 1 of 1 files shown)


docs/developer/changeSpecs/backend-mime_and_downstream.md
```diff
@@ -0,0 +1,32 @@
...
+Propose how integration with downstream postprocessors and tools like SkinnyWMS would be utilized.
+
+# MIME type
+A block instance can currently output multiple types: qubed, raw, none; possibly expanding in the future.
+The raw type can hold images like png, or documents like pdf.
+To determine the underlying type, we currently need to deserialize the raw bytes at the backend, and look up based on class or type what the correct mime type should be.
+
+**Proposal**: extend the Raw type with optional MIME type string field, and set it at the fable compile time.
+1. *Assumption*: we don't expect qubed or future possible types to be ever interpretable by the browser, or if they are, their MIME type to be static.
+2. *Assumption*: we expect MIME type to be determinable at the compile time, ie, without access to the runtime.
```
HCookie, line 12: While this will limit what we can do, I think it's an okay tradeoff

docs/developer/changeSpecs/backend-mime_and_downstream.md
```diff
@@ -0,0 +1,32 @@
...
+Propose output type / mime type contract offered by the backend.
+Propose how integration with downstream postprocessors and tools like SkinnyWMS would be utilized.
+
+# MIME type
+A block instance can currently output multiple types: qubed, raw, none; possibly expanding in the future.
+The raw type can hold images like png, or documents like pdf.
+To determine the underlying type, we currently need to deserialize the raw bytes at the backend, and look up based on class or type what the correct mime type should be.
+
+**Proposal**: extend the Raw type with optional MIME type string field, and set it at the fable compile time.
+1. *Assumption*: we don't expect qubed or future possible types to be ever interpretable by the browser, or if they are, their MIME type to be static.
```
HCookie, line 11: Could be tricky, but I think doable

docs/developer/changeSpecs/backend-mime_and_downstream.md
```diff
@@ -0,0 +1,32 @@
...
+
+The options 2 and 3 are not different from the PoV of the tools or plugins -- either way the correct mime type must be set in advanced in the plugin and thus the set of supported types must be fixed in fiab-core, and we must have a `type->tool` lookup in the backend.
+But they differ in terms of blueprint persistence and reusability, such as:
+* When I submit the job, do I know in advance which postprocessing tools to launch?
+* When I re-launch my job or re-use it to build another job, do I want the same set of tools to be launched afterwards?
+
+Note that the options are _not_ exclusive -- the first one can't be prevented, and we can allow both in-advance and post-hoc launch.
+However, this may muddy the "what happens if I re-launch the job" question -- some of the post-proc tools would get re-launched, others won't.
+Lastly, the Option 2 actually also allows in-advance specification -- the frontend has access to the output type at compile time too, so it can offer the user to launch a tool _but_ remember it itself, not persist it in the backend.
```
HCookie, line 30: I think 2 is a good option, with a potential UI update to allow for launching from a priorly computed output?

tmi: > launching from a priorly computed output?

if you mean priorly like "15 minutes ago" then it should be there from the start

but if you mean like "2 weeks ago", then we would need to persist the url from the RawOutput and possibly check that the location is still valid (because we don't manage it ourselves). The first one is doable, the second one is technically easy but makes it in some scenarios less useful

liefra: From my side, option 2 looks good as well. I like the idea of modelling this using a tools analogy, which the frontend and potentially other clients can use. If the backend maintain a tools registry that clients can connect to in order to retrieve a list of available tools, the returned MIME type could include the tool name and the required parameter signature for a given tool (in this case SkinnyWMS). It may also be useful to include a validation feature to check whether a tool can still be used when someone opens a past job execution.

tmi:
closed since all implemented

PR #422, opened 2026-05-05
Feat/expand artifacts contract

tmi:
A few things happening here:
- we add the is_compatible flag to the artifacts registry, along with a possible detail
- we filter incompatible ones out in the anemoi utils for getting available checkpoints
- we do *not* filter incompatible ones out in the frontend -- we assume the frontend would instead want to gray them out and display the reason
- we future-proof for other types of artifacts, ie, we dont hardcode artifact==ml model checkpoint internally anymore. The anemoi utils api in ecmwf block is unchanged (it filters and casts internally), as is the model detail (we would add new routes if there are new types of artifacts)
- we update the artifacts.json to match the new schema (@HCookie , were you creating these with some automated tool? If yes, I'd be happy to change that too)
- we update the frontend minimalistically, to match the new API -- cc @liefra , here's the seed doc again https://github.com/ecmwf/forecast-in-a-box/blob/2e1661aa7c12a64c9aebf26c700ade6446f2d6df/docs/developer/changeSpecs/backend-artifactsContractExtension.md (not present in the PR anymore, it was just intermediate...)


HCookie (approved):

liefra:
The provided frontend changes are all good to go

PR #428, opened 2026-05-07
[plugin] feat: add GRIB sink block

liefra:
This PR adds a GribSink block: A Sink block that writes a SimpleFieldList to a GRIB file via earthkit-data's file target with append=True. Concurrent writers from cascade fan-out are serialised by an advisory flock on the output file itself so multi-MB messages don't interleave.

Mirrors the existing ZarrSink shape: title/path config, NoOutput on validate, .map(write_grib) on compile. Tests mirror TestZarrSink with EkdSource / EnsembleStatistics / TemporalStatistics fixtures.

_Note_: In its current form, it does not include pass-through of the actual GRIB filename in its output (as we will need later on for the SkinnyWMS integration). It is focused solely on writing the GRIB file.

@tmi , @HCookie : Could you please have a look, especially at the `write_grib` method as it introduces a file lock, which was required in my testing; otherwise, concurrent cascade writes resulted in an inconsistent Grib file state.

Also, please note that I have no idea whether the locking, as currently implemented, would even work in an HPC environment. Could you double-check that as well?

Please amend as required.

### DIFF
--- a/backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/__init__.py
+++ b/backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/__init__.py
@@ -12,13 +12,14 @@ from fiab_core.tools.blocks import QubedBlockBuilder
 from fiab_core.tools.plugins import QubedPluginBuilder
 
 from fiab_plugin_ecmwf.anemoi.blocks import AnemoiSource, AnemoiTransform
-from fiab_plugin_ecmwf.blocks import EkdSource, EnsembleStatistics, MapPlotSink, TemporalStatistics, ZarrSink
+from fiab_plugin_ecmwf.blocks import EkdSource, EnsembleStatistics, GribSink, MapPlotSink, TemporalStatistics, ZarrSink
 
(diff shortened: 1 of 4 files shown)


tmi:
I'm not generally opposed to merging this if it unblocks something such as more skinnyWMS testing, but there are three caveats:
1/ If there is a file lock, it makes no sense to to fan out, it blocks cascade workers needlessly -- instead, we should concatenate to a single dataset and write that as a single writer. The concatenate can be a standalone task with N inputs and 1 outputs (don't we have one such already?), or the write task itself can take N inputs and process them sequentially
2/ I would worry that the flock won't work in the HPC environments, and surely won't work in eg S3 targets
3/ It feels rather odd to implement grib2file in the fiab level, this should probably be in earthkit.workflows itself since its a direct earthkit-data usage, ie, no external dependency

@HCookie thoughts?

liefra:
Note: It’s not crucial for my frontend work to continue working on the skinnyWMS feature. For now, I can use it as a temporary local workaround instead

jinmannwong:
Closing this PR as it has been re-created in https://github.com/ecmwf/forecast-in-a-box/pull/452 (I could not push to this branch).

PR #430, opened 2026-05-07
validation v2

tmi:
Implements the user stories as described in https://github.com/ecmwf/forecast-in-a-box/blob/69a771627d66dcdbcba76a98ae06f3a0c15c720e/docs/developer/changeSpecs/backend-validationV2.md

Implementation follows the guideline https://github.com/ecmwf/forecast-in-a-box/blob/69a771627d66dcdbcba76a98ae06f3a0c15c720e/docs/developer/changeSpecs/backend-validationV2.proposal01.md, and is in detail summarized in https://github.com/ecmwf/forecast-in-a-box/blob/69a771627d66dcdbcba76a98ae06f3a0c15c720e/docs/developer/changeSpecs/backend-validationV2-impl-00-summary.md

Expected frontend is summarized in https://github.com/ecmwf/forecast-in-a-box/blob/69a771627d66dcdbcba76a98ae06f3a0c15c720e/docs/developer/changeSpecs/backend-validationV2-frontendImpact.md, and minimal funcionality-preservation changes were carried out in https://github.com/ecmwf/forecast-in-a-box/pull/430/changes/69a771627d66dcdbcba76a98ae06f3a0c15c720e (cc @liefra)

Leftovers preserved in https://github.com/ecmwf/forecast-in-a-box/blob/03d81cb4151ccf1c4ca4ff5101c9feb644422f63/docs/developer/changeSpecs/backend-validationV2-leftovers.md -- these will be addressed later. Either is nits, or big standalone user stories


docs/developer/changeSpecs/backend-validationV2-impl-00-summary.md
```diff
@@ -0,0 +1,61 @@
...
+- Do not rely on missing configuration values being omitted before task 5.
+- Do not rely on non-empty expansion restrictions before task 7.
+- Preserve compilation as the strict path: missing values, unknown glyphs, malformed glyphs, and type conversion errors must block compilation.
+- Preserve malformed glyph expressions as hard validation errors; only unknown glyph references become soft validation warnings in task 8.
+
+## Concerns and doubts to resolve during review
+
+### FableType syntax compatibility
+
+The proposal mentions `enumClosed[...]` and `enumOpen[...]`, while current plugins use `enum[...]`, `date-iso8601`, `datetime`, `optional[int]`, and list syntax such as `list[int]`. The task breakdown assumes the implementation should choose and document one canonical syntax, with any compatibility aliases handled deliberately. If backward compatibility for existing catalogues is required during the migration window, that should be explicit.
```
tmi: let's resolve this as follows:
1/ for start, migrate plugins to `enumClosed[...]` where they currently use `enum[...]`
2/ migrate plugins to `date` where they use `date-iso8601`. Put the iso into the docstring of the type, similarly for the datetime
3/ use `int` where `optional[int]` is used, dont worry about the implementation inside handling the None
4/ the `list[int]` is legitimate

there is no need for backward compatibility on the catalogue level, but these changes must be documented in the frontend-impact document exactly by the agent working on this task, after the work is done

edit the respective task spec accordingly, then delete this concern

docs/developer/changeSpecs/backend-validationV2-impl-00-summary.md
```diff
@@ -0,0 +1,61 @@
...
+
+### `optional[int]` is not in the proposed type system
+
+`fiab-plugin-ecmwf` currently declares `optional[int]` for `ensemble_members`. The original spec says `None` should not be a normal acceptable value and does not list optional types. The migration task should decide whether this becomes an ordinary `int` with frontend/default behavior, a temporary compatibility alias, or a rejected type.
```
tmi: this should be addressed by the previous concern too -- we dont want to allow optionals. The plugin will have to deal with this inside later, and outward type should be just `int`.

docs/developer/changeSpecs/backend-validationV2-impl-00-summary.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Converted values conflict with current `BlockInstance` typing
+
+`BlockInstance.configuration_values` is currently `dict[str, str]`. After task 4 the backend will pass parsed values such as `int`, `float`, `date`, `datetime`, and lists to plugins. This likely requires widening the internal type, creating an internal converted block model, or using a separate converted configuration mapping. The task should avoid unsafe casts that hide the contract change.
```
tmi: good observation -- lets keep it simple for now and change to dict[str, Any]. This will cause a type checking issue during validation/compilation, as we call the `validate_convert` -- thus make the signature of validate_convert to accept Any, but check inside that its a string and raise TypeError otherwise

docs/developer/changeSpecs/backend-validationV2-impl-00-summary.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Validation vs create/update semantics
+
+The current `/blueprint/create` and `/blueprint/update` routes call validation and reject any returned `block_errors`. Task 5 makes missing configuration values non-errors during validation. That means incomplete blueprints may become saveable through these existing routes unless another strict path is added. This appears aligned with the `prevalidate` user story, but it is a behavior change and should be reviewed explicitly.
```
tmi: this is indeed aligned. Do the following -- to the leftovers document, put "add a metadata flag to the blueprint persistence which would signify whether its ready for compilation, ie, no missing config options or glyphs". It should not be addressed in this work, ie, put in the worker spec that this is accepted behaviour change

docs/developer/changeSpecs/backend-validationV2-impl-00-summary.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Expand restriction shape at the HTTP boundary
+
+The core plugin method can return local `BlockExpansion` values with local `BlockFactoryId`, but `/blueprint/expand` currently returns `PluginBlockFactoryId` values so the frontend can identify the owning plugin. Task 6 needs a route-level serialized shape that keeps plugin identity and adds restrictions without making the response ambiguous.
```
tmi: correct. Make it explicit in the worker spec that this needs to be handled correctly. In the contract in fiab-core, the BlockExpansion would contain the BlockFactoryId. But the backend route response (in the domain.blueprint.service would contain the PluginBlockFactoryId, and the service does the cast. This pattern is _already_ used in the regular expand functionality, thus the worker is to replicate this pattern

PR #443, opened 2026-05-18
Use stored run output mime

tmi:
We used to derive the mime type from the actual object that cascade returns as the task output

We change that to utilize the mime type that has been derived during compilation

cc @liefra -- this should allow you to remove the workarounds you were forced to introduce. However, I have *not* verified that all blocks in the plugins set the mime type correctly during compilation. I'd say we best merge this, remove workarounds, and if we detect a mime type issue, we go and fix the plugin in the first place.

### DIFF
--- a/backend/src/forecastbox/domain/run/cascade.py
+++ b/backend/src/forecastbox/domain/run/cascade.py
@@ -20,6 +20,7 @@ import xarray as xr
 from cascade.gateway.api import JobSpec, ResultRetrievalResponse, SubmitJobRequest, SubmitJobResponse, decoded_result
 from cascade.gateway.client import request_response
 from cascade.low.core import JobInstance, JobInstanceRich, TaskId
+from cascade.low.func import Either
 from fiab_core.fable import BlockInstanceId
 from pydantic import Field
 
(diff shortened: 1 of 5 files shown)


backend/src/forecastbox/domain/run/cascade.py
```diff
@@ -42,47 +43,80 @@ class ExecutionSpecification(FiabBaseModel):
...
 
-def encode_result(result: ResultRetrievalResponse) -> tuple[bytes, str]:
-    """Converts cascade Result response to bytes+mime"""
+def encode_result(result: ResultRetrievalResponse, mime_type: str) -> Either[bytes, str]:  # ty: ignore[invalid-type-arguments]
+    """Converts cascade Result response to bytes for a known mime type."""
+    # TODO this function should change as follows:
+    # - the preferred way is that cascade tasks return simply bytes, so thats what we'll start with
+    # - the (tuple) thing should be completely dropped, it increases memory pressure and we dont really
+    #   need that safety. First, drop it from all plugins, then from here
+    # - all the other options should be deleted, first ensured they are not used in plugins
```
tmi, line 52: @HCookie @jinmannwong can I have you review this comment? It deals with what the *runtime* of a block should return -- what I'm saying here that it is always *bytes*, in the form the browser would understand

ie, if its an image, it should not return `ekp.Figure`, but `figure.save(buf)` (thats what the plugin currently does, except for still having that tuple-hack in place). If its a grib to be saved to the disk, the plugin would return the path_to_grib.encode('ascii'), etc.

if you agree with this strategy, I'll go through the plugins and remove what I'll find, then start simplifying this function -- so I'm asking you more about "are you happy with this going forward?" and "are you aware of any existing place where this will cause issues?"

liefra:
Yes, you can merge and I'll do a follow up and remove/recheck the current mime type sniffing workarounds

backend/src/forecastbox/domain/run/cascade.py
HCookie: What about the instances when grib or files should be downloaded, would that return bytes, or a have a file lookup?

tmi: > What about the instances when grib or files should be downloaded, would that return bytes, or a have a file lookup?

if it should be downloadable, I would always return `bytes`. Not sure if there is a meaningful case where we would want to both store to disk _and_ expose for a direct download -- I mean, why download something which is already on disk? It is true that these two disks *can* be different, ie, cascade-on-cluster writes to S3, not your local disk where you would download a thing on. But for that I would probably develop a new Lens which is basically rsync2local -- which may be the "file lookup" you had in mind? Lets keep that as a future option, and generally stick to "plots and images are downloadable and thus return bytes, grib files and other binary data go to disk and return a url", and solve the "cascade running elsewhere" later

liefra: > downloadable and thus return bytes

for proper handling on the client side it is okay to provide bytes, but we should also have the correct mime type as well to signal the browser what applications/previewers are relevant. e.g. `image/png`, `application/pdf`, `image/svg+xml`, `application/netcdf` , `application/x-grib`, `application/x-grib2` (if we allow direct download of grib/netcdf)

PR #445, opened 2026-05-19
[plugin]: declare standard output MIME for Map Plot blocks

liefra:
This is a follow-up plugin-related PR ensuring that the `MapPlotSink` correctly set the MIME type using `mime_type`.

ISSUE: `MapPlotSink.validate()` set its output MIME on `RawOutput.type_fqn` instead of `.mime_type` — the field compile reads — so Map Plot outputs recorded `application/octet-stream`. Now that the run-output endpoint trusts the stored
MIME, that mismatched the runtime's `image/*` result and returned 500; plots stopped rendering.

FIX: Declares it on `.mime_type` with standard types (`application/pdf`, `image/svg+xml`), shared between `blocks.py` and `runtime/plots.py` via a `FORMAT_MIME_TYPES` map. Adds tests for both sides.

### Contributor Declaration

### DIFF
--- a/backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
+++ b/backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
@@ -28,6 +28,7 @@ from fiab_core.tools.blocks import Product, Sink, Source
 from qubed import Qube
 
 from .qubed_utils import axes, contains, coxpand, dimensions
+from .runtime.plots import FORMAT_MIME_TYPES
 
 SOURCE = ConfigurationOptionId("source")
 DATE = ConfigurationOptionId("date")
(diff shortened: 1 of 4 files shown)


HCookie (approved):

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
```diff
@@ -28,6 +28,7 @@
 from qubed import Qube
 
 from .qubed_utils import axes, contains, coxpand, dimensions
+from .runtime.plots import FORMAT_MIME_TYPES
```
tmi, line 31: ftr -- this would have been a problem, we cant have block factories import from runtime code; the import relationship must be the other way around. Sadly we dont have in place any means of preventing/detecting this at the moment

tmi:
closing in favour of https://github.com/ecmwf/forecast-in-a-box/pull/446 which I started working on in parallel. The fixing of the mime in the plots block is similar there (except the problematic runtime import here), but we simplify the return type to be bytes only -- as well as handle that in the test plugin, and on the backend itself

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
liefra: Thanks for pointing this out and clarifying!

PR #449, opened 2026-05-19
Feat/compilation detail

tmi:
This exposes the underlying cascade subgraph to the frontend, for better visualization

It is done via a new endpoint, /getCompilationDetail, which requires a runId and optionally accepts a block id -- without blockId, retrieves the whole graph, when blockId is given it filters only to the respective block

it returns nodes of the cascade graph, as a list, each element containing a display name and ids of the parents. Note that if you give a block id in the request to filter and that block id is not a source, then you will get elements in that list whose parents are _not_ in the response

the display name is currently the long verbose name with a hash. There are a few things we can think about:
 - do we want to store some better display_name in the metadata during compilation?
 - do we want to put there some *type* into the metadata? Like "compute node", "auxiliary manipulation node", "data fetch node", "data store node", etc? We could derive like color or icon from it (the original graph viz also wasn't homo)
 - do we want to put in some "grouping" information, to provide hints to visualization like "all these nodes are a single parallel computation group"?
 
:point_up: cc @corentincarton @jinmannwong @HCookie 

there are a few internal inefficiencies, and the RunProgress response does _not_ yet contain the granular level of task-id, I will add this later -- but this should already enable some integrations, cc @liefra

### DIFF
--- a/backend/src/forecastbox/domain/run/compile.py
+++ b/backend/src/forecastbox/domain/run/compile.py
@@ -10,7 +10,6 @@
 """Compilation of a BlueprintBuilder into an ExecutionSpecification."""
 
 from datetime import datetime
-from typing import cast
 
 from cascade.low.core import DatasetId, TaskId
 from cascade.low.func import assert_never
@@ -26,7 +25,7 @@ from forecastbox.domain.glyphs.intrinsic import AvailableIntrinsicGlyphs, get_va
 from forecastbox.domain.glyphs.resolution import merge_glyph_values, resolve_configurations
 from forecastbox.domain.plugin.manager import PluginManager
 from forecastbox.domain.run.cascade import ExecutionSpecification, RawCascadeJob, RunOutputCharacteristic, RunOutputs
-from forecastbox.domain.run.detail import CompilationDetail, TaskDetail, fluentNode_to_detail
+from forecastbox.domain.run.detail import CompilationDetail, TaskDetail, _fluentName_to_taskId, fluentNode_to_detail
 from forecastbox.domain.run.types import RunId
 from forecastbox.utility.graph import topological_order
 from forecastbox.utility.time import value_dt2str
@@ -87,8 +86,6 @@ def compile_builder(
     plugins = PluginManager.plugins
     action_lookup = {}
     block_outputs: dict[BlockInstanceId, BlockInstanceOutput] = {}
-    # Maps any produced cascade TaskId (node.name) → originating BlockInstanceId.
-    task_to_block: dict[TaskId, BlockInstanceId] = {}
     # Maps any produced cascade TaskId → TaskDetail.
     task_detail: dict[TaskId, TaskDetail] = {}
     # Maps sink block ids to mime type used in RunOutputs (only relevant for external outputs).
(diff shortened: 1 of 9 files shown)


backend/src/forecastbox/domain/run/detail.py
```diff
@@ -0,0 +1,75 @@
...
+
+    block: BlockInstanceId
+    display_name: str
+    parents: list[TaskId]
+
+
+class CompilationDetail(FiabBaseModel):
+    """Detail produced by compilation of a BlueprintBuilder."""
+
+    task_to_block: dict[TaskId, BlockInstanceId]
```
tmi: this is now redundant -- refactor existing uses of this field to derive it from the `task_detail` below

backend/src/forecastbox/domain/run/detail.py
```diff
@@ -0,0 +1,75 @@
...
+    task_detail: dict[TaskId, TaskDetail] = {}
+
+
+def fluentNode_to_detail(node: Node, block: BlockInstanceId) -> TaskDetail:
```
tmi: ideally return a tuple[TaskId, TaskDetail] here. Additionally, create a `_fluentName_to_taskId` function here, which would encapsulate the `TaskId(<name>)` -- because thats something that may change in the future. Note how the current codebase in compile.py has a comment `# TODO hack` related to these invocation, thats precisely why we want to centralize it

backend/src/forecastbox/domain/run/detail.py
```diff
@@ -0,0 +1,75 @@
...
+    """Convert an earthkit.workflows Node to a TaskDetail.
+
+    Parameters
+    ----------
```
tmi: keep docstrings brief, not this numpy-style breakdown of parameters

backend/src/forecastbox/domain/run/compile.py
```diff
@@ -88,6 +89,8 @@ def compile_builder(
     block_outputs: dict[BlockInstanceId, BlockInstanceOutput] = {}
     # Maps any produced cascade TaskId (node.name) → originating BlockInstanceId.
     task_to_block: dict[TaskId, BlockInstanceId] = {}
```
tmi: the task_to_block is not needed anymore, is it?

backend/src/forecastbox/domain/run/service.py
```diff
@@ -244,8 +244,9 @@ def _build(
         task_to_block: dict[TaskId, BlockInstanceId] | None = None
         if detailed_report:
             try:
-                task_to_block = cast(dict[TaskId, BlockInstanceId], get_memcache(run_id, dict))
-            except (KeyError, TypeError):
+                compilation_detail = retrieve_compilation_detail(run_id)
+                task_to_block = compilation_detail.task_to_block
+            except (CompilationDetailNotFound, CompilationDetailCorrupted):
                 detailed_report = False
                 warning_error = "unable to provide completed/planned tasks"
```
tmi: extend the warning error with `repr` of the exception

backend/src/forecastbox/routes/run.py
```diff
@@ -287,6 +303,46 @@ async def get_run(
...
+    auth_context: AuthContext = Depends(get_auth_context),
+) -> CompilationDetailResponse:
+    """Return task-level compilation detail for a run.
+
+    If ``block_id`` is provided, only tasks belonging to that block are returned.
+    Returns 404 if the run is not found or if no compilation detail is available
+    (e.g. the run has not yet been submitted, or the cache entry has expired).
+    """
+    try:
+        await db.get_run(spec.run_id, spec.attempt_count, auth_context=auth_context)
```
tmi: why do we attempt to get the run? For the access rights check? Lets ignore that for this purpose, we will later handle it when we move from memcache to persistent layer

PR #450, opened 2026-05-20
[frontend] backend sync (#443/#447) + builder splice & undo/redo

liefra:
Aligns the frontend with backend changes from #443 to #447. 

While testing the alignment two builder-UX features came up — drop/popover edge splicing and an undo/redo stack — and are included here.

### Backend sync (#443, #446, #447)

- **API error detail.** `getJobResult`, `getJobResultHead`, and `downloadJobLogs` now surface the FastAPI `detail` body via a new `apiErrorFromResponse` helper instead of dropping it for `response.statusText`. Backend #443's detail-rich 500s ("Result mime lookup failed: …") previously rendered as bare "Internal Server Error" in toasts.
- **Mocks.** `/run/outputContent` MSW handler reads `exec.outputs[datasetId].mime_type` instead of always returning PNG. Seed data on `job-completed-001` now spans `image/png`, `application/pdf`, and `image/svg+xml` so PdfThumbnail and the image-vector adapter are exercised in dev. A `noStoredOutputsExecution` fixture drives the new "no recorded outputs yet" 500 path.
- **Restrictions plumbing.** `BlockExpansion.restrictions` now flows into `BlockValidationState.expansionRestrictions` and overlays the option's `value_type` in `ConfigPanel`. The new `selectVariable.variable` option (#447) renders as a `ClosedEnum` dropdown of the upstream's params. The value-type parser also accepts the unquoted form emitted by `FableType.serialize()` (e.g. `enumClosed[2t,msl]`) alongside the catalogue's quoted form.
- **Viewers.** `ImageThumbnail`, `ImageViewer`, and the copy-image action stop hardcoding `image/png` for raster outputs — they trust `item.mimeType` when the adapter advertises it and fall back to the adapter's primary mime otherwise. Latent today (no first-party sink emits JPEG/WebP yet) but unblocked for future sinks.
- **Housekeeping.** Dropped the unread `contentType` return from `getJobResult`; refreshed adapter comments to cite the standard mimes the backend now emits.

### Builder UX

- **Drop-splice.** Dropping a transform/product block on an existing edge inserts inline (`source → new → consumer`) instead of leaving the new block as a side branch. Mirrored across the `+` popover (`AddNodeButton`) and the palette drag-drop (`useSidebarBlockDrop`). Sinks and zero-input factories skip the splice. Input-handle drops perform the symmetric splice — the input's prior parent becomes the new block's upstream.
- **Undo/Redo on the configure page.** Bounded `past`/`future` snapshot stack of `fable` state with same-key coalescing for typing bursts and a transaction wrapper so multi-action UI flows (drop, popover-add with splice) collapse to a single undo step. Cmd/Ctrl+Z and Cmd/Ctrl+Shift+Z (Ctrl+Y on non-Mac) drive it via `@tanstack/react-hotkeys`; two icon buttons next to "Save Config" expose the same actions to mouse users. History resets on `setFable`/`newFable`; selection clears on restore when the selected block isn't in the restored snapshot.

### Contributor Declaration


liefra:
Rebased the PR onto 9cb759a7 and merged the `MultiSelectField` work into the `EnumListField`, as introduced in 9cb759a7. 

@corentincarton: Please note that my changes to the `EnumListField` make use of the shadcn [combobox](https://ui.shadcn.com/docs/components/base/combobox), as this is a better out-of-the-box component for supporting multi-selection functionality. Added also filtering.

Additionally, I removed the variable/glyph toggle option for `EnumListField` fields and the related flickering effect that occurred when selecting an item from the popup/combobox.

Feel free to stash commits.

See also the screenshot of the final result:

Recording 2026-05-21 at 11-42-02

corentincarton:
Thanks @liefra, your multiple input selector looks much better indeed! @tmi, I'm merging these changes as I prefer not to leave the main branch with my hacky frontend code.

PR #453, opened 2026-05-20
Add SSH tunnel utility

tmi:
1/ Introduces a utility for setting up and managing ssh connexions
2/ Refactors the gateway management to distinguish between RemoteUrl, RemoteTunnel, and LocalProcess:
  - RemoteUrl does not manage gateway at all, no logs (yet) or restarts supported
    provide full url of cascade gateway `tcp://<hostname>:<port>` and set spawn: False
  - RemoteTunnel supports spawning/restart, but no logs (yet)
    provide url `ssh://[<username>@]<hostname>` and set spawn: True
  - LocalProcess supports spawning/restart and logs
    provide url `tcp://localhost` and set spawn: True

Changes to config:
  - local gateway does not require port in its url -- we always choose a random one
  - dropped unused config options related to cascade (log size, venv dir)
  - introduced a new config for logging base (use like /tmp/cascadeLogs or /home/<hpcLogin>/fiabLogs). Default behaviour is the original one
 
Notes:
  - artifacts are _not_ supported correctly
  - beware on hpc -- if you ssh to hpc in parallel to observe/troubleshoot, you may be on a different node! You will not `ps` see the process, you will not see the same `/tmp`, etc. I was using `fiab__cascade__cascade_logging_base=/home/ecme7737/fl fiab__cascade__cascade_url="ssh://hpc-login" FIAB_ROOT=.fiab uv run --no-sync python -m forecastbox.entrypoint.main` to execute it
  - we *assume* `uv` to be installed and available, but we always run `uv run --with earthkit-workflows`, meaning you dont need to install it, but you also override it to eg editable via config
  - slurm submissions are *not* supported yet -- so basically dont run anything :)

### DIFF
--- a/backend/src/forecastbox/domain/gateway/__init__.py
+++ b/backend/src/forecastbox/domain/gateway/__init__.py
@@ -1,0 +1,16 @@
+# (C) Copyright 2024- ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""
+Manages the Gateway domain -- process lifecycle and connection URL used by
+other domains to execute and inspect workflow jobs.
+
+Depends on utility config and Cascade runtime bindings.
+Depended on by Run domain and gateway/status routes.
+"""
--- a/backend/src/forecastbox/domain/gateway/service.py
+++ b/backend/src/forecastbox/domain/gateway/service.py
@@ -1,0 +1,158 @@
+# (C) Copyright 2024- ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Gateway lifecycle and URL management for the backend."""
+
+import logging
+import os
+import threading
+from dataclasses import dataclass
+from multiprocessing.process import BaseProcess
+from tempfile import TemporaryDirectory
+
+from cascade.deployment.logging import LoggingConfig
+from cascade.executor import platform
+from cascade.gateway import api, client
+from cascade.gateway.server import main_enp
+
+from forecastbox.domain.gateway.exceptions import (
+    GatewayAlreadyRunning,
+    GatewayExited,
+    GatewayManagerNotInitialized,
+    GatewayNotRunning,
+    GatewayNotStarted,
+)
+from forecastbox.utility.config import StatusMessage, config
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True, eq=True, slots=True)
+class GatewayProcess:
+    logs_directory: TemporaryDirectory
+    process: BaseProcess
+
+
+@dataclass(eq=False, slots=True)
+class GatewayConnectionManager:
+    lock: threading.Lock
+    gateway_process: GatewayProcess | None = None
+
+
+_manager: GatewayConnectionManager | None = None
+
+
+def _require_manager() -> GatewayConnectionManager:
+    if _manager is None:
+        raise GatewayManagerNotInitialized("Gateway manager is not initialized")
+    return _manager
+
+
+def init_manager() -> None:
+    global _manager
+    if _manager is not None:
+        raise ValueError("Gateway manager is already initialized")
+    _manager = GatewayConnectionManager(lock=threading.Lock())
+
+
+def _reset_manager() -> None:
+    global _manager
+    _manager = None
+
+
+def launch_cascade(cascade_url: str, log_base: str | None, max_concurrent_jobs: int | None) -> None:
+    logging_config = LoggingConfig(path_base=log_base, formatter="line")
+    try:
+        main_enp(url=cascade_url, loggingConfig=logging_config, max_concurrent_jobs=max_concurrent_jobs)
+    except KeyboardInterrupt:
+        pass
+
+
+def launch_gateway() -> None:
+    manager = _require_manager()
+    with manager.lock:
+        if manager.gateway_process is not None:
+            raise GatewayAlreadyRunning("Process already running.")
+        logs_directory = TemporaryDirectory(prefix="fiabLogs")
+        logger.debug(f"logging base is at {logs_directory.name}")
+        logs_base = None if os.getenv("FIAB_LOGSTDOUT", "nay") == "yea" else logs_directory.name + os.sep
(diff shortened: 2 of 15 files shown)


backend/src/forecastbox/utility/tunnel.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def _claim_free_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        sock.bind(("127.0.0.1", 0))
+        return int(sock.getsockname()[1])
+
+
+def _claim_remote_port() -> int:
+    return _REMOTE_PORT_MIN + secrets.randbelow(_REMOTE_PORT_MAX - _REMOTE_PORT_MIN + 1)
```
tmi: is there a reason we use `secrets` for random number, as opposed to `random`?

backend/src/forecastbox/utility/tunnel.py
```diff
@@ -0,0 +1,218 @@
...
+    host: str
+    control_path: str
+    local_port: int
+    remote_port: int
+
+    @property
+    def cascade_url(self) -> str:
+        """Return the local Cascade URL exposed through the tunnel."""
+
+        return f"tcp://localhost:{self.local_port}"
```
tmi, line 106: lets not name this `cascade_url` -- remember this is utility module, ie, it should be possible to use it for other remote commands from this codebase in the future. `as_local_url` is a better name to the method (and dont make it a `@property`, we dont like that here)

backend/src/forecastbox/utility/tunnel.py
```diff
@@ -0,0 +1,218 @@
...
+    def discard(self, handle: ConnectionHandle) -> None:
+        with self._lock:
+            self._handles.discard(handle)
+
+    def shutdown(self) -> None:
+        with self._lock:
+            handles = tuple(self._handles)
+        for handle in handles:
+            stop(handle)
+
```
tmi, line 129: Do we need the contains, len and iter methods at this stage? I guess no -- lets remove them, we want to keep the codebase without bloat

backend/src/forecastbox/domain/gateway/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""
+Manages the Gateway domain -- process lifecycle and connection URL used by
+other domains to execute and inspect workflow jobs.
+
+Depends on utility config and Cascade runtime bindings.
+Depended on by Run domain and gateway/status routes.
```
tmi, line 15: well written -- but ideally update the domain/run/init as well accordingly

backend/src/forecastbox/domain/gateway/service.py
```diff
@@ -0,0 +1,158 @@
...
+    if _manager is None:
+        raise GatewayManagerNotInitialized("Gateway manager is not initialized")
+    return _manager
+
+
+def init_manager() -> None:
+    global _manager
+    if _manager is not None:
+        raise ValueError("Gateway manager is already initialized")
+    _manager = GatewayConnectionManager(lock=threading.Lock())
```
tmi: thinking of it, we should not need this method, because the constructor is actually not doing anything. Lets drop the init and _require methods and initialize at import time (we do it with other managers as well). This also removes the need for the "manager not initialized" exception

backend/src/forecastbox/domain/gateway/service.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def _reset_manager() -> None:
+    global _manager
+    _manager = None
```
tmi: not needed, we will never reset the manager

backend/src/forecastbox/domain/gateway/service.py
```diff
@@ -0,0 +1,236 @@
...
+    if parsed_url.hostname is None or parsed_url.port is None:
+        raise ValueError(f"Invalid cascade_url for remote tunnel mode: {config.cascade.cascade_url!r}")
+    host = f"{parsed_url.username}@{parsed_url.hostname}" if parsed_url.username else parsed_url.hostname
+    return host, parsed_url.port
+
+
+def _random_local_gateway_url() -> str:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as sock:
+        sock.bind(("127.0.0.1", 0))
+        port = int(sock.getsockname()[1])
```
tmi: use the claim_free_port from utilities/tunnel.py here -- ideally remove the leading `_` from that name, it can be a public method

backend/src/forecastbox/domain/gateway/service.py
```diff
@@ -0,0 +1,236 @@
...
+    return f"tcp://localhost:{port}"
+
+
+GatewayConnectionManager.gateway_connection = _initial_connection()
```
tmi: move this call into the class body -- I don't want initialization to be spread around

backend/src/forecastbox/domain/gateway/service.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+def launch_cascade(cascade_url: str, log_base: str | None, max_concurrent_jobs: int | None) -> None:
```
tmi: lets rename to _local_process_entrypoint -- should not be public, should be clearly bound to local process

backend/src/forecastbox/domain/gateway/service.py
```diff
@@ -0,0 +1,236 @@
...
+            host, remote_port = _remote_tunnel_target_and_port()
+            handle = tunnel.setup(host=host, remote_port=remote_port)
+            remote_gateway_url = f"tcp://localhost:{handle.remote_port}"
+            cmd = ["uv", "run", "python", "-m", "cascade.gateway", "--url", remote_gateway_url]
+            if max_concurrent_jobs is not None:
+                cmd.extend(["--max_concurrent_jobs", str(max_concurrent_jobs)])
+            tunnel.execute(handle, cmd)
+            GatewayConnectionManager.gateway_connection = RemoteTunnel(handle=handle)
+            return
+        raise NotImplementedError("RemoteUrl gateway cannot be launched by backend")
```
tmi: dont write as `if & return`, instead write as `if .., elif .., elif .., else: assert_never` -- search for the assert_never idiom usage here. You will need to be explicit about matching the RemoteUrl instance.

This will make it easier when we add in the future another subtype of GatewayConnection -- type system will tell us about all the places. Apply this pattern to the remaining isinstance matchings on GatewayConnection in this file.

backend/src/forecastbox/domain/gateway/service.py
```diff
@@ -0,0 +1,236 @@
...
+    if gateway_connection is None or isinstance(gateway_connection, RemoteUrl):
+        return
+    if isinstance(gateway_connection, LocalProcess):
+        if gateway_connection.process.exitcode is None:
+            try:
+                stop_gateway()
+            except GatewayNotRunning:
+                pass
+        else:
+            GatewayConnectionManager.gateway_connection = None
```
tmi: we are in shutdown so should be ok, but lets be careful and lock before setting to None. Same for the case of RemoteTunnel, handled a few lines below

PR #454, opened 2026-05-20
[plugin] Extend earthkit-data source functionality

jinmannwong:
### Description
Extends earthkit-data source block to allow retrieval of control forecast and pressure level/wave parameters

### Contributor Declaration


backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/datasets/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        return qube
+
+    def is_member_zero(self, datacube: dict) -> bool:
+        if not self.member_zero:
+            return False
+        return all([self.member_zero[key] in datacube[key] for key in self.member_zero])
+
+
+def load_datasets() -> dict[str, ForecastDataset]:
+    dataset_dir = os.path.join(LOCALDIR, "configs")
```
tmi: nit: I think this will not work in 100% cases, only in like 99%. It works for editable installs and for wheel installs into a venv, but there are also installs from like zipfile and whatnot, so there is some canonical way of getting the pkg resources. I will fix it later and tag you there, no need to block this PR with it

tmi (approved):

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/datasets/utils.py
```diff
@@ -0,0 +1,51 @@
...
+    datasets = {}
```
HCookie: Can we use https://docs.python.org/3.13/library/importlib.resources.html here instead?

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/qubed_utils.py
```diff
@@ -210,3 +211,22 @@ def _contains_axis_values(key: str, values: Iterable[Any]) -> bool:
...
+    -----
+    >>> output = QubedOutput(dataqube=Qube.from_datacube({
+    ...     'param': ['2t', 'tp'],
+    ...     'time': [0, 1, 2],
+    ... }))
+    >>> selection = select(output, {'param': '2t'})
+    >>> axes(selection)
+    {'param': {'2t'}, 'time': {0, 1, 2}}
+    """
+    return QubedOutput(dataqube=qselect(qube.dataqube, selection))
```
HCookie: I think `select` exists on the qube object itself

```suggestion
    return QubedOutput(qube.dataqube.select(selection))
```

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/datasets/utils.py
jinmannwong: Ah, ok, thanks.

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/datasets/configs/ifs-ens.yaml
HCookie, line 1: How did you make this? Is it worth adding a script to automate it?

jinmannwong: I've added the script for creating the dataset configuration files from the opendata index. This is more restrictive than what is available in mars, but it will us a dataset that can is available in both.

tmi (approved):

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
```diff
@@ -85,6 +65,8 @@
     "svg": "image/svg+xml",
 }
 
+FORECAST_DATASETS = load_datasets()
+
 
 class EkdSource(Source):
```
corentincarton: Shouldn't we rename this block and its description into something more forecast oriented rather than technology oriented?
Maybe ArchivedForecastSource, ECMWFForecastSource or something?
We should edit the description accordingly.

corentincarton (changes requested):
Looks really good! @jinmannwong, we should just discuss of a new name for the earthkit source and then we can merge. It would be great to brainstorm a bit on how we could also create a ECMWFRDExperimentSource or something like that :)

corentincarton (approved):

PR #463, opened 2026-05-28
Cascade-Gateway Improvements

tmi:
# Description
1/ general config changes for more clarity and coherence (backwards incompatible, see below for "migration guide")
2/ bump of ewk & corresponding api adaptation
3/ fix unit and integration tests to not be affected by .env files and .fiab homes
4/ extend the config and job submission to support slurm/sshCluster submissions

# Config Migration Guide
1. Field name changes:
   - config.api → config.backend (BackendAPISettings → BackendSettings)
   - config.product (artifact_stores, plugin_stores, plugins) → config.external (ExternalServicesSettings)
   - config.product.model_repository moved to config.external.model_repository

2. Gateway settings restructuring:
   - Replaced single spawn_gateway boolean with discriminated union of three gateway types
   - LocalGateway for local testing (no cascade_url needed, defaults to tcp://localhost)
   - RemoteGateway for remote SSH-based gateway management
   - UnmanagedGateway for externally managed gateway instances
   - Use isinstance() checks instead of spawn_gateway boolean
   - Gateway-specific startup parameters (max_concurrent_jobs, cascade_logging_base) are now in startup_params

3. Cascade constraints:
   - Moved max_hosts, default_hosts, default_workers_per_host, max_workers_per_host to config.cascade.constraints

### DIFF
--- a/backend/src/forecastbox/domain/blueprint/cascade.py
+++ b/backend/src/forecastbox/domain/blueprint/cascade.py
@@ -7,7 +7,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-"""Cascade execution data models: environment specification."""
+from typing import Literal
 
 from fiab_core.artifacts import CompositeArtifactId
 from pydantic import Field, PositiveInt
@@ -17,6 +17,7 @@ from forecastbox.utility.pydantic import FiabBaseModel
 
 class EnvironmentSpecification(FiabBaseModel):
     # NOTE warning -- this class is used by the web api. Be careful about changes here
+    cascade_infra: Literal["localProcess", "slurm", "sshCluster"] = Field(default="localProcess")
     hosts: PositiveInt | None = Field(default=None)
     workers_per_host: PositiveInt | None = Field(default=None)
     environment_variables: dict[str, str] = Field(default_factory=dict)
--- a/backend/src/forecastbox/domain/blueprint/cascade.py
+++ b/backend/src/forecastbox/domain/blueprint/cascade.py
@@ -7,6 +7,8 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+"""Cascade execution data models: environment specification."""
+
 from typing import Literal
 
 from fiab_core.artifacts import CompositeArtifactId
--- a/backend/src/forecastbox/domain/gateway/service.py
+++ b/backend/src/forecastbox/domain/gateway/service.py
@@ -31,7 +31,7 @@ from forecastbox.domain.gateway.exceptions import (
     GatewayNotStarted,
 )
 from forecastbox.utility import tunnel
-from forecastbox.utility.config import StatusMessage, config
+from forecastbox.utility.config import LocalGateway, RemoteGateway, StatusMessage, UnmanagedGateway, config
 
 logger = logging.getLogger(__name__)
 _LOCAL_HOSTS = {"localhost", "127.0.0.1", "::1"}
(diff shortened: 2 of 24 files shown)


backend/src/forecastbox/utility/config.py
```diff
@@ -235,15 +234,38 @@ def validate_runtime(self) -> list[str]:
                 errors.append(f"ssh:// data_path must have an absolute path: {self.data_path}")
         else:
             errors.append(f"unsupported scheme in data_path (use file:// or ssh://): {self.data_path}")
-        if not _validate_url(self.model_repository):
-            errors.append(f"not an url: model_repository={self.model_repository}")
```
tmi, line 239: you should move this to the runtime validate of ExternalServicesSettings

backend/src/forecastbox/domain/gateway/service.py
```diff
@@ -59,17 +59,18 @@ class RemoteUrl:
 
 
 def config2gatewayMethod() -> GatewayConnectionType:
```
tmi: I think this method is not needed anymore, and neither is the Literal it return -- you should be able to replace all of its invocations by isinstance check (which you typically had to do anyway at the callsite)

this method has one extra which would be lost by isinstance check -- it checks the scheme in case of RemoteGateway. Lets move that to the runtime validate method of the RemoteGateway class

backend/src/forecastbox/domain/run/cascade.py
```diff
@@ -22,12 +22,46 @@
...
 from forecastbox.domain.blueprint.cascade import EnvironmentSpecification
 from forecastbox.domain.gateway.service import get_gateway_url
-from forecastbox.utility.config import config
+from forecastbox.utility.config import UnmanagedGateway, config
 from forecastbox.utility.pydantic import FiabBaseModel
 
 logger = logging.getLogger(__name__)
 
 
+def _select_cascade_infra(environment: EnvironmentSpecification) -> Literal["localProcess", "slurm", "sshCluster"]:
```
tmi: don't repeat the definition of the Literal -- introduce it as a type alias `CascadeInfrastructureType = Literal[...` in the utility/config.py and then use that

backend/src/forecastbox/domain/blueprint/cascade.py
```diff
@@ -7,7 +7,7 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
-"""Cascade execution data models: environment specification."""
```
tmi, line 10: dont delete this docstring

backend/src/forecastbox/domain/run/cascade.py
```diff
@@ -22,12 +22,46 @@
...
+        ssh_cluster_spec = gateway.startup_params.ssh_cluster_spec
+        if ssh_cluster_spec is None:
+            raise ValueError("sshCluster submissions require cascade.gateway.startup_params.ssh_cluster_spec")
+        return SshCluster(
+            controller_url=ssh_cluster_spec.controller_url,
+            worker_urls=ssh_cluster_spec.worker_urls,
+            workers_per_host=workers_per_host,
+        )
+
+    raise ValueError(f"Unsupported cascade infrastructure: {requested_infra}")
```
tmi: convert the `if, if, if, raise` pattern into `if, elif, elif, else: assert_never` -- look for usages of `assert_never` in the codebase

backend/src/forecastbox/domain/run/cascade.py
```diff
@@ -22,12 +22,46 @@
...
+    if requested_infra == "slurm":
+        if not isinstance(gateway, UnmanagedGateway):
+            if not gateway.startup_params.shared_path:
+                raise ValueError("slurm submissions with managed gateway require cascade.gateway.startup_params.shared_path")
+        return SlurmCluster(workers_per_host=workers_per_host, hosts=hosts)
+
+    if requested_infra == "sshCluster":
+        ssh_cluster_spec = gateway.startup_params.ssh_cluster_spec
+        if ssh_cluster_spec is None:
+            raise ValueError("sshCluster submissions require cascade.gateway.startup_params.ssh_cluster_spec")
```
tmi: amend this and previous message in the ValueError with `require specifying config value in`)

backend/src/forecastbox/utility/config.py
```diff
@@ -246,11 +246,20 @@ def validate_runtime(self) -> list[str]:
...
+    worker_urls: list[str]
+
+
 class GatewayStartupParams(FiabBaseModel):
     max_concurrent_jobs: int | None = 1
     """If more jobs submitted at a given time, all but this many wait in a queue"""
     cascade_logging_base: str | None = None
     """Where to store logs of cascade gw and jobs. Use eg /home/<user>/fiabLogs or /tmp/fiabLogs"""
+    shared_path: str | None = None
+    """Shared filesystem path visible to all workers, required for managed Slurm submissions."""
```
tmi: remove the managed

backend/src/forecastbox/utility/config.py
```diff
@@ -275,9 +284,12 @@ class UnmanagedGateway(FiabBaseModel):
     gateway_type: Literal["unmanaged"]
     cascade_url: str
     """Base URL for the Cascade API, eg tcp://<hostname>:<port>"""
+    startup_params: GatewayStartupParams = Field(default_factory=GatewayStartupParams)
```
tmi: UnmanagedGateway must *not* have startup params, because we dont start it. I guess you include it because the ssh cluster needs access to those -- lets then raise NotImplementedError in that case instead

PR #478, opened 2026-06-02
Implement c-tag publishing flow

tmi:
1/ adds a new gittag method to artifacts store config
2/ said gittag retrieves all tags of the form cX.Y.Z.d from the repo, filters for those which share X.Y.Z with current fiab-core version, and retrieves the artifacts.json for the highest of them
3/ makes the fiab-core release verify that current artifacts.json is compatible with it
4/ adds a new gh action that pushes d+1 for the given cX.Y.Z, verifying there is no other change except artifacts.json, and that even changed it still parses

### DIFF
--- a/.github/workflows/cd-plugins.yaml
+++ b/.github/workflows/cd-plugins.yaml
@@ -8,7 +8,7 @@ on:
       tag:
         description: Explicit tag. Use cX.Y.Z, make sure its free on pypi
         type: string
-        required: true
+        required: false
       testpypi:
         description: Whether to upload to testpypi instead of pypi.
         type: boolean
@@ -32,12 +32,31 @@ jobs:
           if [ "${{ inputs.testpypi }}" = "true" ] ; then
             set -x
           fi
+          git fetch --tags origin
+
           TAG_PREF="c"
           TAG_FULL="${{ inputs.tag }}"
-          case "$TAG_FULL" in
-            c*) ;;
-            *) echo "expected tag in the form cX.Y.Z" >&2; exit 1 ;;
-          esac
+          if [ -n "$TAG_FULL" ] ; then
+            echo "$TAG_FULL" | grep -Eq '^c[0-9]+\.[0-9]+\.[0-9]+$' || {
+              echo "expected tag in the form cX.Y.Z" >&2
+              exit 1
+            }
+          else
+            LATEST_BASE=$(
+              git tag -l 'c*' \
+              | sed -nE 's/^c([0-9]+\.[0-9]+\.[0-9]+)(\.[0-9]+)?$/\1/p' \
+              | sort -V \
+              | tail -n1
+            )
+            if [ -z "$LATEST_BASE" ] ; then
+              LATEST_BASE="0.0.0"
+            fi
+            IFS='.' read -r MAJOR MINOR PATCH <<< "$LATEST_BASE"
+            NEW_PATCH=$((PATCH + 1))
+            TAG_FULL="${TAG_PREF}${MAJOR}.${MINOR}.${NEW_PATCH}"
+            echo "determined tag = $TAG_FULL"
+          fi
+
           TAG="${TAG_FULL}.0"
           export SETUPTOOLS_SCM_PRETEND_VERSION=${TAG_FULL#$TAG_PREF}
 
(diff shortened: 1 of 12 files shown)


.github/workflows/cd-plugins.yaml
```diff
@@ -6,7 +6,7 @@ on:
   workflow_dispatch:
     inputs:
       tag:
-        description: Explicit tag. Use vX.Y.Z, make sure its free on pypi
+        description: Explicit tag. Use cX.Y.Z, make sure its free on pypi
         type: string
         required: true
```
tmi: the autoderivation was not implemented -- ie, the tag should not be required. When not given, it should query past tags, strip the last number to obtain X.Y.Z, pick the numerically highest, add 0.0.1, and convert back to cX.Y.Z.0 form

backend/packages/fiab-core/scripts/parse_artifacts_catalog.py
```diff
@@ -0,0 +1,31 @@
...
+    parser.add_argument("path", type=Path)
+    args = parser.parse_args(argv)
+
+    data = args.path.read_text()
+    dict(parse_json(ArtifactStoreId("validation"), data, lambda _checkpoint: (True, None)))
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
```
tmi: too complicated, just call main() here, and main can be annotated with return None. We dont need to return 0 in case of success, we just want to raise in case of failure

backend/src/forecastbox/utility/git.py
```diff
@@ -0,0 +1,57 @@
...
+"""Helpers for reading repository tags from GitHub."""
+
+import re
+from collections.abc import Iterator
+
+import httpx
+
+_GITHUB_OWNER = "ecmwf"
+_GITHUB_REPO = "forecast-in-a-box"
+_TAG_RE = re.compile(r"^c(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(?:\.(?P<build>\d+))?$")
```
tmi: this hardcodes the C -- but we could keep this universal. Lets rebrand as "PREFIXED_VERSION_TAG_RE", make the first character any letter (so `[a-z]`). Update the docstring of `get_highest_tag` as well as the string in ValueError

backend/src/forecastbox/domain/artifact/catalog.py
```diff
@@ -34,27 +45,19 @@ def get_artifacts_catalog(artifact_stores_config: ArtifactStoresConfig) -> Artif
...
+                raw = fetch_content(_tagged_url(store_config.url, tag), client).decode("utf-8")
             else:
                 assert_never(store_config.method)
 
+            for composite_id, artifact in parse_json(
+                store_id,
+                raw,
+                lambda store_info: get_model_checkpoint_compatibility(store_info, platform_info),
+            ):
+                catalog[composite_id] = artifact
```
tmi: you should not need to materialize the whole catalog like this, ie, just map the iterator, then feed it to `pmap`

backend/src/forecastbox/domain/artifact/catalog.py
```diff
@@ -34,27 +45,19 @@ def get_artifacts_catalog(artifact_stores_config: ArtifactStoresConfig) -> Artif
...
+                logger.debug(f"Loaded artifact {composite_id} from store {store_id}")
```
tmi: no need to debug log this as we'll construct on the fly

backend/src/forecastbox/utility/config.py
```diff
@@ -155,7 +155,7 @@ def _default_plugin_stores() -> PluginStoresConfig:
 
 class ArtifactStoreConfig(FiabBaseModel):
     url: str
-    method: Literal["file"]
+    method: Literal["file", "gittag"]
```
tmi, line 158: add a validation to this class that if the method is gittag, _then_ the url must be a string which contains ${TAG} substring. Then change the `_tagged_url` method in artifacts.py to just format it with the tag (maybe you will have to go with %TAG, {TAG}, whatever makes `format` happy)

it would be nice to use a t-string here but we want to support older python versions too

PR #481, opened 2026-06-04
[plugins-ecmwf]: Upgrade to ekw-anemoi 0.7.0

HCookie:
### Description
This pull request introduces support for handling multiple datasets in Anemoi model checkpoints by allowing model input and output qubes to represent collections of datasets, and adds a new block for selecting a specific dataset from model outputs. It also refactors utility functions to be more flexible, improves input configuration logic, and updates dependencies.

**Support for multiple datasets and dataset selection:**

* Added the `AnemoiDatasetSelect` block to allow users to select a specific dataset from the output of an Anemoi model, including validation and configuration options. (`backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/blocks.py`, `backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/__init__.py`) [[1]](diffhunk://#diff-986656ee6c60fe2c1f2feee364b670c2ef6b09140ab590b8bb590f55d337825eR275-R306) [[2]](diffhunk://#diff-c9403c8584dedb84b6580759eda4f5ec58297f78b31308c92b1a01293a0445ccL14-R14) [[3]](diffhunk://#diff-c9403c8584dedb84b6580759eda4f5ec58297f78b31308c92b1a01293a0445ccR52)
* Updated `CheckpointArtifact` methods to handle both single and multiple dataset qubes, including combining multiple qubes into a single qube with a dataset dimension and expanding steps for each dataset. (`backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/utils.py`)

**Refactoring and generalization of utility functions:**

* Refactored `collapse`, `expand`, `axes`, and related functions in `qubed_utils.py` to support both `Qube` and `QubedOutput` types via a decorator, improving flexibility and code reuse. (`backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/qubed_utils.py`) [[1]](diffhunk://#diff-f061390536fc5c2584536e708d019a039b8ee981facc632eeeb7a69689754f5cL30-R63) [[2]](diffhunk://#diff-f061390536fc5c2584536e708d019a039b8ee981facc632eeeb7a69689754f5cL46-R81) [[3]](diffhunk://#diff-f061390536fc5c2584536e708d019a039b8ee981facc632eeeb7a69689754f5cL67-R98) [[4]](diffhunk://#diff-f061390536fc5c2584536e708d019a039b8ee981facc632eeeb7a69689754f5cL97-R136)

**Improvements to block validation and compilation:**

* Updated `AnemoiSource`, `AnemoiInputSource`, and `AnemoiTransform` blocks to use the new logic for handling nested dataset qubes and to always return `QubedOutput` for consistency. (`backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/blocks.py`) [[1]](diffhunk://#diff-986656ee6c60fe2c1f2feee364b670c2ef6b09140ab590b8bb590f55d337825eL148-R157) [[2]](diffhunk://#diff-986656ee6c60fe2c1f2feee364b670c2ef6b09140ab590b8bb590f55d337825eL199-R205) [[3]](diffhunk://#diff-986656ee6c60fe2c1f2feee364b670c2ef6b09140ab590b8bb590f55d337825eL237-R256)

**Enhancements to input configuration logic:**

* Improved the `get_input_configuration` method to correctly handle cutout input configurations for nested models, avoid mutating input arguments, and return configurations in the expected format. (`backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/utils.py`) [[1]](diffhunk://#diff-be733c2f434eb142a49bf382f3c170b7eee6b58136db9a6d953c5987b4b733e4R139-L124) [[2]](diffhunk://#diff-be733c2f434eb142a49bf382f3c170b7eee6b58136db9a6d953c5987b4b733e4R152-R172)

**Dependency and compatibility updates:**

* Bumped the `earthkit-workflows-anemoi` dependency to `>=0.7.0` for compatibility with the new dataset features. (`backend/packages/fiab-plugin-ecmwf/pyproject.toml`)
* Made the package version extraction script more robust by skipping missing dependencies. (`backend/packages/fiab-plugin-ecmwf/scripts/get_metadata_from_anemoi.py`)

### Contributor Declaration


backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/blocks.py
```diff
@@ -265,3 +272,35 @@ def compile(  # type:ignore[invalid-argument] # semigroup
...
+    configuration_options: dict[ConfigurationOptionId, BlockConfigurationOption] = {
+        DATASET: BlockConfigurationOption(
+            title="Dataset",
+            description="Name of the dataset to select from the Anemoi model output",
+            value_type="str",
+        ),
+    }
+
+    def intersect(self, other: QubedOutput) -> bool:
+        return "dataset" in axes(other)
```
tmi: this feels like an odd design -- so on one hand, the block is about selecting from outputs of Anemoi models, but here we just trigger on "dataset" being in the input. Why do we need a dedicated select based on previous block type? And if we do, shouldn't we then be more careful on the intersection?

couldn't we instead make the original select block more generic? Like a generic qubed selector, where you say "this dimension, this attribute"?

tmi:
CI failed on type check, but this looks like agent-implemented -- are your agents reading the agents md? It leads to the instruction of running just val prior to commit, but apparently they aren't

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/blocks.py
HCookie: Sure, it is quite generic, perhaps a better generic select? But then it might not be clear for a user that they need to do so...

HCookie:
@tmi Tried the fix with Copilot button, it did not

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/blocks.py
HCookie: Resolved in [d152018](https://github.com/ecmwf/forecast-in-a-box/pull/481/commits/d152018c7e426f5fc9bd32c2d277cd4694bfecbb)

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
```diff
@@ -411,6 +411,73 @@ def intersect(self, other: QubedOutput) -> bool:
...
+        block_id: BlockInstanceId,
+        block: BlockInstance,
+    ) -> Either[Action, Error]:  # type:ignore[invalid-argument] # semigroup
+        dimension = block.config_as_str(ConfigurationOptionId("dimension"))
+        values = block.config_as_list(ConfigurationOptionId("values"), str, allow_empty=False)
+        input_task = block.input_ids["dataset"]
+        selected = inputs[input_task].select({dimension: values if len(values) > 1 else values[0]})
+        return Either.ok(selected)
+
+    def restrictions(self, other: QubedOutput) -> ConfigurationOptionRestriction:
```
corentincarton: @HCookie, while I fully agree with the purpose of this block, it doesn't work in practice as you get all the coordinates in the values configuration, which is very confusing. @tmi may kill me for suggesting this, but maybe there's a way to add the BlockInstance of the current block as input to the restriction method and call it every time we change a configuration option? We could then have a unique select block.

tmi: > but maybe there's a way to add the BlockInstance of the current block as input to the restriction method

@corentincarton  what do you mean by a block instance of the current block? You get the `self` which is the class of an instance you are _possibly_ constructing, but the instance does not exist yet, ie, the user has not created it in the graph at all. You also get the `other` which is the input, but you cannot get the full class that created it, because that may come from another plugin

maybe the right compromise is the following -- have a single generic class for Select, with dimension being a param to the constructor, and use it to create multiple block factories, like "SelectVariables" and "SelectDatasets" and "SelectWhatNot", each with a fixed dimension, and each being offered in the `expand` only when said dimension is in the input.

my original comment to @HCookie , which led to this refactoring, was concerned with the internal code structure, and a desire to keep it coherent. On the UI level, we can offer it expanded -- I have no concern there myself

there is an alternative solution -- the user would see a single select block, and would in a first step select dimension, in the second step the value, so a sort of gradual expansion. But only the first round of providing the `Restrictions` would be handled by the `restrictions` method, the subsequent rounds would be handled by `validate`, which already accepts the partially constructed BlockInstance (I guess this is what you asked for, didn't you?). The `validate` currently returns `BlockInstanceOutput`, and we could change it to return `BlockInstanceOutput | ConfigurationOptionRestriction`, in the sense of "either its completely specified and configured and thus provides an output, or its incomplete and thus provides restrictions to completion". But note that it puts quite a burden on the plugin author! Nowadays, you need to reason either "there are _no_ config values, only input blocks" in restrictions *or* "there are _all_ config values, and you only decide good/bad". This alternative proposal would expand it to "there can be any subset of config values, and you need to provide best effort on validation-restriction any time". I would not mind it but then I won't be writing the plugins

> @tmi may kill me for suggesting this

as said in the previous paragraph, I can do something worse, I can implement it :D But then the others may not exactly like the added complexity

HCookie: I think what Corentin meant was that the restrictions function gets the configuration as it stands from the ui?
That way it could be a generic block without the validate?

HCookie: > maybe the right compromise is the following -- have a single generic class for Select

I think we have that already, as an alternative right now, I can create one for dataset?

tmi: > I think we have that already

oh right, https://github.com/ecmwf/forecast-in-a-box/blob/main/backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/__init__.py#L42, then I'm even more puzzled why we introduced the AnemoiSelect...

tmi: > I think what Corentin meant was that the restrictions function gets the configuration as it stands from the ui?

but by the time `restrictions` is called, there are no configuration ui values at all -- because thats part of the `expand` call, ie, _before_ the block is even picked by the user. Note that `restrictions` is actually a return value of the `expand` call, its only on the PluginBuilder level that these two are separated, see https://github.com/ecmwf/forecast-in-a-box/blob/main/backend/packages/fiab-core/src/fiab_core/tools/plugins.py#L69

and the possible change is to broaden the `restrictions` signature with `instance`, give there an empty one when its called by `expand`, and additionally call it in `validate`. But you would need to reason about whether it should be called before/after validate, and in the after case, whether it should be called if validate fails. The problem is we currently fail validate on a missing value, which would make this useless, because if there is no missing value then there can be no further restrictions. But on the other hand, if we called this every time, then you could not rely on the present values being valid, so you would have to repeat every validation code inside the `restrictions`. The best solution would be to broaden the validation to distinguish between "soft" and "hard" fails, as we discussed already some time ago, but we don't have that yet. Or do the restrictions derivation within the validate call -- that would make most sense to me, but I guess you prefer the validate and restriction separated, as you did in expand?

HCookie:
Can we try to get this merged this week?

PR #494, opened 2026-06-11
Feat/fiabcore compat check

tmi:
backwards incompatible change due to database schema expansion

1/ Stores the current fiab-core major version into each Blueprint database entity when created
2/ When retrieving a blueprint (via get/list route), we include a new tag CoreVersionMismatch -- presumably for the frontend to display with a warning (cc @liefra)
3/ We change the schema of the blueprint entity's `tags` to be list[{key, value|None}] instead of list[str]. A minimalistic change has been included to preserve current frontend behaviour, ie, converting `tag: str` to `key=tag, value=None`, in https://github.com/ecmwf/forecast-in-a-box/commit/45c5e4cc1bfb8f15cfe7f39d2a4be27b5c3b717a (cc @liefra ). My suggestion would be to display the value in the tooltip when hovered over the key, or smth like that. The CoreVersionMismatch tag contains the f"{blueprint's core version} != {current core version}" string. It is intended as a soft warning, not a hard error which prevents running a job from that blueprint or working with it further
4/ When installing a plugin, we include a guard with the current version of fiab-core to prevent accidental core upgrades
5/ The update_plugin now accepts a version to install to. It is not checked whether this version is compatbile, ie, we do allow explicit breach if the user asks for it. When version is not given, we install the newest one that is compatible
6/ There is a new endpoint, plugins/versions, which returns a list of compatible versions for a given plugin

### DIFF
--- a/backend/packages/fiab-plugin-ecmwf/justfile
+++ b/backend/packages/fiab-plugin-ecmwf/justfile
@@ -1,5 +1,5 @@
 val:
-    uv run ty check
+    uv run ty check src tests
     uv run pytest ./tests/unit && uv run --extra runtime pytest ./tests/runtime
 
 build:
--- a/backend/src/forecastbox/domain/blueprint/db.py
+++ b/backend/src/forecastbox/domain/blueprint/db.py
@@ -39,7 +39,7 @@ async def upsert_blueprint(
     builder: dict | None = None,
     display_name: str | None = None,
     display_description: str | None = None,
-    tags: list[str] | None = None,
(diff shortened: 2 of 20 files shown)


liefra:
Frontend changes in [45c5e4c](https://github.com/ecmwf/forecast-in-a-box/pull/494/commits/45c5e4cc1bfb8f15cfe7f39d2a4be27b5c3b717a) are good to be merged. Once merged will make further adjustments in the UI to ensure these tags can be properly displayed.

PR #495, opened 2026-06-11
[frontend/backend]: in-app WMS viewer for stored sink outputs via SkinnyWMS lenses

liefra:
## Summary

Adds an end-to-end "inspect your forecast on a map" flow for GRIB sink outputs: the run detail page lists every disk-written sink output, and each row can launch a SkinnyWMS lens to open an in-app OpenLayers WMS viewer or copy a standard WMS GetCapabilities URL for external clients (QGIS, ArcGIS, …). 

Following the review comment, the backend diff reduced to only provide `SKINNYWMS_CORS_ORIGINS` for spawned lens instances. Everything else is frontend. 

The previously discussed backend WMS proxy is deliberately **not** part of this PR (deferred per discussion) — the browser talks to lens ports directly, which works for local/single-host deployments.

## How it works

1. GribSink already streams its glyph-resolved output directory as the task's output payload (mime `text/plain; fiab-format=gribdir`, `GRIB_MIME` in the ecmwf plugin). That payload is the single source of truth: the frontend derives one "Stored outputs" row per sink block from the run's own outputs and fetches the tiny payload to learn the directory.
2. "Open" starts a SkinnyWMS lens on that directory and mounts the in-app map viewer (parameter browser with pressure-level grouping, multi-layer stack with opacity/reorder, time slider with animation and optional prefetch, legends, basemap switcher incl. SkinnyWMS native decoration layers, PNG export). The copy action copies the GetCapabilities URL for external WMS clients. Lens lifecycle is explicit: a running server shows a status badge with its port and a Stop control on the row; "Active lenses" lists all running instances.
3. The directory served by the lens is the parent directory of the sink's configured output path. There is currently no enforcement of per-run isolation — it is up to the user to configure a path whose directory is dedicated to the run, so that the GRIB files in it belong to that run only. The intrinsic glyphs make this straightforward, e.g. `path: /data/output/${runId}__${attemptCount}/forecast_[shortName].grib2` puts every run's files into their own directory; the AIFS Ensemble Dataset preset ships this pattern as its default. With a shared directory (e.g. `path: /tmp/out_[shortName].grib2`), the WMS server would expose whatever else that directory contains. A plugin-enforced per-run output directory would remove this responsibility from the user.

## Frontend

- `WmsViewer` (OpenLayers `ImageWMS`), `wms-capabilities.ts` (WMS 1.3.0 parser: leaf layers, decoration split, bbox, ISO8601 time-interval expansion, parameter grouping/humanising), `StoredOutputsCard`, `ActiveLensesCard`, lens API client/hooks/types.
- New output adapter for the GribSink marker mime: output cards show a GRIB chip instead of falling back to the generic binary adapter (the payload is a directory path, so the card offers no download).
- Graceful degradation: where the optional SkinnyWMS package is not installed (linux/aarch64 per #496), `/lens/supported` omits it and the frontend disables the lens actions with an inline explanation instead of offering actions that can only fail.
- New deps: `ol`, `ol-mapbox-style`. CSP: external basemap hosts added; `localhost:*` connect/img sources are injected **in dev builds only** (vite transform) — prod CSP gains no localhost entries.
- MSW lens handlers (stateful start → poll → running → stop, plus a skinny-not-installed mode) so mock-mode dev and tests cover the full lifecycle.

## Not included / limitations

- No backend proxy route: lens traffic goes directly to `http://<backend-host>:<port>`, so the viewer requires the browser to reach the lens ports (true for local and single-host deployments). The proxy can be added later without frontend rework — only `buildLensBaseUrl` changes.
- Per-run isolation of the served directory is the user's/configuration's responsibility (see "How it works" item3)

## Testing

- Frontend: 40 new tests (capabilities parser, lens URL builders, StoredOutputsCard/ActiveLensesCard integration against the MSW lens handlers incl. start→running→stop lifecycle, payload-derived rows, and the SkinnyWMS-unavailable state); full suite 1171 unit/integration + e2e + typecheck/lint/build green.
- Manually verified against a live backend with real AIFS GRIB runs: open viewer, layer/time interaction, copy URL, stop lifecycle, run-private directory serving.

### Contributor Declaration


backend/src/forecastbox/domain/run/cascade.py
```diff
@@ -85,6 +85,9 @@ class RunOutputCharacteristic(FiabBaseModel):
 
 class RunOutputs(FiabBaseModel):
     outputs: dict[TaskId, RunOutputCharacteristic]
+    # block_id → filesystem path for sinks that write to disk (Zarr, GRIB, NetCDF…).
+    # Populated at compile time by walking sink blocks' `path`/`dir` config keys.
+    stored_outputs: dict[BlockInstanceId, str] = {}
```
liefra: Channel through which the frontend learns where sinks wrote

tmi: I'm very puzzled by this -- I thought that the output itself, which holds the url of the path, is sufficient for this. It contains precisely this value, doesn't it? Why do we then need this?

backend/src/forecastbox/domain/run/compile.py
```diff
@@ -94,6 +94,9 @@ def compile_builder(
     # Maps sink block ids to mime type used in RunOutputs (only relevant for external outputs).
     block_to_mime: dict[BlockInstanceId, str] = {}
     sink_tasks: set[TaskId] = set()
+    # Sinks that write to a filesystem path (Zarr, GRIB, NetCDF…) — block_id → path
+    stored_outputs: dict[BlockInstanceId, str] = {}
+    _STORED_PATH_KEYS = ("path", "dir")
```
tmi: this looks very heuristical, I don't think we want to rely on such naming convention. If the block wants to declare any manner of output, it must come from the BlockOutput class, not from inspecting the config! Because you cannot know if there is any transformation of the path inside the plugin which handles the configuration key

backend/src/forecastbox/domain/lens/manager.py
```diff
@@ -77,6 +85,33 @@ def _compute_status(instance: LensInstance) -> LensInstanceDetail:
...
+
+    - an existing single file → ``[local_path]``
+    - a GribSink file-pattern (``[shortName]``) → every glob match
+    - an existing directory → ``[]`` (caller serves the directory directly)
+    """
+    if os.path.isfile(local_path):
+        return [local_path]
+    if os.path.isdir(local_path):
+        return []
+    return sorted(glob.glob(_pattern_to_glob(local_path)))
```
tmi: I'm moderately puzzled by this -- I think we should have the plugin always create and output a directory, hence this whole machinery would not be needed, right?

liefra: Yes, that is correct and if the plugin always create and outputs writes into a dedicated per-run directory then the staging/symlink machinery here disappears and additionally the changes in `cascade.py` and `compile.py` can be dropped entirely and the frontend derives the rows from the existing outputs. This would be a cleaner approach.

liefra:
If we agree to make changes to the plugin to always create and outputs writes into a dedicated per-run directory, then I guess this this PR should wait for these changes. Will put the PR in draft for now.

liefra:
Following @tmi's comments, I reworked the branch and the backend changes. As most of the backend changes are not actually required if the user chooses a unique folder location for a GRIB file (e.g. `/tmp/${runId}__${attemptCount}/[shortName].grib2` ). The backend implementation I had previously added became therefore obsolete and has been removed in this update. 

The backend diff reduced to only provide `SKINNYWMS_CORS_ORIGINS` for spawned lens instances.

We can then discuss (and implement into another PR) if it would be better if plugins to always create and outputs writes into a dedicated per-run directory.

Force-pushed and the frontend now also checks `/lens/supported` and disables the WMS actions with an inline explanation on deployments without SkinnyWMS (based on changes in  #496).

Removed draft status as well.

PR #496, opened 2026-06-11
[backend] make skinny optional, exclude on linux aarch

tmi:
cc @HCookie turned out to be easy -- pylock seems to handle conditional dependencies nicely

thus on aarch linux machine, it will simply not install it, and it will then not show in the supported lenses

once it becomes available, you can simply pip install it post hoc to the fiab venv and it starts working -- and at the same time we would remove the discriminator from the pyproject

cc @EduardRosert pls ping us once it becomes available, thx :)


backend/src/forecastbox/routes/lens.py
```diff
@@ -44,6 +44,13 @@
     responses={404: {"description": "Not found"}},
 )
 
+try:
+    import skinnywms
```
HCookie, line 48: Don't want to use `importlib`?

HCookie (approved):

backend/src/forecastbox/routes/lens.py
tmi: why would we?

PR #498, opened 2026-06-11
[frontend/backend] PC: high-level presets

HCookie:
## Summary

Introduces a **high-level preset** system: curated, versioned forecast templates with a small parameter surface that non-expert users can launch from the dashboard, while admins can author and publish presets via a CRUD page. Utilises glyphs and blueprints as the primary drivers of the new functionality.

### Backend
- New `domain/preset` package: `HighLevelPreset` ORM + Pydantic models, optimistic-locking DB layer, instantiation service, dynamic `glyph[catalogue:…]` value-type resolver, and a 5-preset seed (`quick-temperature-map`, `regional-surface-forecast`, `aifs-ensemble-to-grib`, `global-ensemble-statistics`, `blank-canvas`).
- New `/api/v1/presets` router: `list`, `get`, `create`, `update`, `delete`, `publish`, `instantiate` with admin gating and `PresetAccessDenied → 403` mapping.
- Plugin manager: `wait_until_ready` helper used from `/get` via `asyncio.to_thread` so cold-start catalogue resolution does not block the event loop.

### Plugin (ecmwf)
- Block `metadata.environment` fix and lazy `as_catalogue()` value-type overrides so presets can resolve options at request time.

### Frontend
- Preset API client, TanStack Query hooks, Zod types, new `presets` i18n namespace.
- Dashboard preset gallery and getting-started cards with per-difficulty launch behaviour (beginner = instant run, intermediate = wizard then run, advanced = wizard then open in editor).
- `PresetWizardDialog` with `previewMode` so the admin form can render a preview without calling `/instantiate`.
- Admin CRUD page at `/admin/presets` with live preview.
- New routes: `/presets`, `/presets/gallery`, `/presets/mine`, `/admin/presets`. `/configure` accepts `?hlPreset=` and redirects legacy `?preset=` slugs.

### Dashboard – Preset Gallery
image

### Preset Wizard Dialog
image

image

### Admin – Preset Form + Live Preview
image

image


corentincarton:
Thanks @HCookie, that's a step in the right direction! Shouldn't we move the list of presets to the plugin? I think we should avoid as much as we can plugin related code in the backend and frontend (even if it's not a direct dependency here).

HCookie:
Reopened as a proof of concept, to be assessed for viability and if this is a useful concept for users. It may be mined for smaller sub PRs or closed and replaced by other more well made code

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
HCookie, line 1: Implemented to address missing dependencies in the operational forecast block

backend/src/forecastbox/domain/preset/value_type_resolver.py
HCookie, line 1: Implemented to enable presets to stay synced with available artefacts or enums as defined by plugins.

HCookie:
## Weave plan
[high-level-presets.md](https://github.com/user-attachments/files/28948539/high-level-presets.md)

PR #508, opened 2026-06-16
Feat/pkg update dt

tmi:
The plugin route now returns field `update_datetime` formatted as datetime, instead of `update_date`. This is to address a sort of frontend bug which originally took date and converted to utc midnight, leading to odd times being displayed. The frontend is simplified to treat the given datetime as is

Overall makes backend more consistent wrt how it reports dates and times

cc @liefra -- as always, the frontend impl is pure prompt, feel free to refuse/rewrite

### DIFF
--- a/backend/src/forecastbox/domain/plugin/manager.py
+++ b/backend/src/forecastbox/domain/plugin/manager.py
@@ -43,7 +43,7 @@ from pyrsistent.typing import PMap
 from forecastbox.domain.plugin.compatibility import install_plugin_compatibly
 from forecastbox.utility.concurrent import delayed_thread, timed_acquire
 from forecastbox.utility.config import PluginSettings, PluginsSettings, config, config_edit_lock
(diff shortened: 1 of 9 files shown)


liefra:
@tmi: Do you think it would be useful/doable to include always the timezone server-side to ensure the frontend can align always correctly, as currently the frontend only works correctly when the backend is started with UTC+0 timezone. 

E.g: "2026-06-16T14:30:00Z" or "2026-06-16T14:30:00+00:00"

Currently there is e.g. also no timezone in `/api/v1/experiment/operational/scheduler/current_time`

tmi:
> Do you think it would be useful/doable to include always the timezone server-side

@liefra sure, let me try that in another PR -- in the meantime, is this mergeable from your POV?

liefra:
@tmi : Yes, good to merge from my side. Thanks

PR #509, opened 2026-06-16
Feat/utc

tmi:
Complete migration of all datetimes to UTC, including:
1/ descriptive routes (plugin update datetime, current scheduling time, glyph example value)
2/ glyph resolution (start_time)
3/ database entries for created_at, updated_at
4/ unified supported formats in jinja parsing and in datetime type parsing for fable configuration blocks -- they now all support the same 4 options. Two of them are *not* tz aware, so we will probably drop them? 

Unexpected were the lack of sqlite-sqlalchemy support for timezones, and formatting differences between 3.11 and 3.12+ python versions

Includes frontend update:
1/ updates stale comments about frontend being tz-naive to frontend being tz-aware, even though code itself apparently needed no change
2/ updates the mocks to match what the backend is now returning
cc @liefra , as always, feel free to update

### DIFF
--- a/backend/src/forecastbox/domain/admin.py
+++ b/backend/src/forecastbox/domain/admin.py
@@ -22,6 +22,7 @@ from pathlib import Path
 import httpx
 
 from forecastbox.utility.config import fiab_home
+from forecastbox.utility.time import current_time
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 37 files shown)


liefra:
@tmi: frontend wise, I'll give it a manual review and proper check once it is merged and good on the backend side. So from my side, frontend parts are okay as they are right now.

PR #513, opened 2026-06-19
fix: strip timezone from datetime inputs in AnemoiBuilder and sources

HCookie:
### Description
Strip timezone from datetime inputs in AnemoiBuilder and sources

### Contributor Declaration

### DIFF
--- a/backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/blocks.py
+++ b/backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/blocks.py
@@ -48,6 +48,12 @@ INPUT_SOURCE = ConfigurationOptionId("input_source")
 BASE_TIME = ConfigurationOptionId("base_time")
 
 
+def strip_timezone(dt: DATE) -> DATE:
+    if dt.tzinfo is not None:
+        return dt.astimezone(tz=None).replace(tzinfo=None)
+    return dt
+
+
 class AnemoiBuilder:
     """Utility to build an Inference from an Anemoi checkpoint, for use in both Source and Transform blocks"""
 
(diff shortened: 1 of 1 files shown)


backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/anemoi/blocks.py
```diff
@@ -48,6 +48,12 @@
 BASE_TIME = ConfigurationOptionId("base_time")
 
 
+def strip_timezone(dt: DATE) -> DATE:
+    if dt.tzinfo is not None:
+        return dt.astimezone(tz=None).replace(tzinfo=None)
```
tmi: this converts to local timezone, not to the utc -- I thought we want utc, no?

HCookie: so as utc then replace with none?

tmi: well that depends on what the user means / what we want

tmi: lets say there is a usecase where we have a workflow where we take "the most recent midnight", then going to utc is the right thing

tmi: but if the user is confused about the system configuration, has it in local time, and put's there the actual utc time (like 00:00:00), then they are screwed...

tmi: but yeah, `astimezone(tz=dt.utc)` is probably the safest/most reasonable

tmi: and then the .replace tzinfo None, thats a must in any case to make the comparison in ekd work

tmi (approved):

PR #526, opened 2026-06-29
feat(fiab-core): add BlueprintTemplate entity and expose on Plugin co…

tmi:
feat(plugin): plugin-owned blueprint templates

Plugins can now ship blueprint templates -- partial, ready-to-customise graphs that are ingested into the blueprints table at install time and made available to all users as starting points.

### fiab-core contract

- New `BlueprintTemplate` entity exposing the public subset of a blueprint builder: `display_name`, `display_description`, `blocks`, `environment` (only `environment_variables`), `local_glyphs`, `example_values`, and `example_glyphs`. Tags and infrastructure fields are intentionally excluded.
- `SelfPluginId` sentinel so plugin code can reference its own factories without knowing its composite ID at authoring time; the backend substitutes the real ID at install time. **Note**: this is a hack. I will later remove the whole PluginId business from the plugin's PoV
- `Plugin.blueprint_templates` field (defaults to empty tuple, fully backwards compatible).
- Four test templates added to `fiab-plugin-test` covering the core scenarios: fixed values, glyph substitution, exclusion, glyph remapping, and deliberate validation failure.

### Backend persistence (plugin install state)

- New `plugin_state` DB table (one row per plugin) storing installed version, timestamp, install error, per-plugin settings (`excluded_templates`, `glyph_remapping`, `enabled`), and per-template validation errors.
- Plugin version and install timestamp are no longer held in memory; they are written to the DB at install time and read back by `GET /plugins/status`. Install errors now survive app restarts.
- Plugin install is guarded by the existing `PluginManager` lock, keeping it atomic.

### Template ingestion

- On every install or re-ingest, each non-excluded template is converted to a `BlueprintBuilder`, glyph names are rewritten per the stored remapping, and the result is validated with `validate_expand(validate_only=True)` after overlaying `example_values`/`example_glyphs`. Templates that fail validation are skipped and their errors are recorded in `plugin_state.template_errors`, which surfaces in the status endpoint.
- Upsert semantics keyed on `(created_by, display_name)`: the same `blueprint_id` is reused across plugin updates so that past runs and user-derived blueprints remain valid. Excluded templates are soft-deleted.

### Settings route

- `POST /plugins/settings` (admin only) persists `excluded_templates` and/or `glyph_remapping` with partial-update semantics (`null` means "leave unchanged", empty list/dict means "clear"), then immediately triggers a re-ingest without reinstalling the pip package.
- toggling `modifyEnabled` to disabled causes all the templates to be deleted as well
- the plugin details route now additionally returns excluded templates and glyph remapping

### Glyph remapping

- `remap_glyph_names` (in `domain/glyphs/resolution`) performs a single non-recursive pass over `${...}` expressions, renaming only AST-level variable identifiers at word boundaries -- filter and global names are untouched.
- `remap_builder_glyphs` (in `domain/blueprint/service`) applies this to every configuration value and to both keys and values of `local_glyphs`.

### Config

- the `enabled` settings for plugin got removed, because it is now on the DB level. This is a **breaking** change -- configs with this field will refuse to load.

## Notes

- `status_full()` became `async`; both call sites in the plugins router updated accordingly.
- There is a known circular dependency between the `plugin` and `blueprint` domains, mitigated with lazy imports and documented in both module docstrings. It will be resolved in a future refactoring towards an event-driven model.
- Filtering the blueprint list by source/plugin is deferred to a subsequent PR.

### DIFF
--- a/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
+++ b/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
@@ -18,10 +18,8 @@ from fiab_core.fable import (
     ConfigurationOptionRestriction,
     NoOutput,
     PluginBlockFactoryId,
-    PluginCompositeId,
-    PluginId,
-    PluginStoreId,
     RawOutput,
+    SelfPluginId,
 )
 from fiab_core.plugin import BlockValidation, Error, Plugin
 from fiab_core.types import FableType
@@ -198,26 +196,34 @@ def compiler(lookup: ActionLookup, instance: BlockInstance) -> Either[Action, Er
 plugin = lambda: Plugin(catalogue=catalogue(), validator=validator, expander=expander, compiler=compiler)
 
 
-_SELF_PLUGIN_ID = PluginCompositeId(store=PluginStoreId("__self__"), local=PluginId("__self__"))
+def _make_source_text_block(text: str) -> BlockInstance:
+    return BlockInstance(
+        factory_id=PluginBlockFactoryId(plugin=SelfPluginId, factory=BlockFactoryId("source_text")),
+        configuration_values={TEXT: text} if text else {},
+        input_ids={},
+    )
 
-_BASIC_BLOCK_ID = BlockInstanceId("text_source")
+
+_BLOCK_FIXED = BlockInstanceId("text_fixed")
+_BLOCK_GLYPHS = BlockInstanceId("text_glyphs")
+_BLOCK_EXAMPLE = BlockInstanceId("text_example")
 
 _testBasic = BlueprintTemplate(
     display_name="testBasic",
-    display_description="A minimal test template with a single source_text block.",
+    display_description="A minimal test template demonstrating fixed values, glyph substitution, and example values.",
     blocks={
-        _BASIC_BLOCK_ID: BlockInstance(
-            factory_id=PluginBlockFactoryId(
-                plugin=_SELF_PLUGIN_ID,
-                factory=BlockFactoryId("source_text"),
-            ),
-            configuration_values={TEXT: "{{ example_text }}"},
-            input_ids={},
-        ),
+        # Fixed: configuration value is a literal string -- no glyphs, no example override needed.
+        _BLOCK_FIXED: _make_source_text_block("fixed text"),
+        # Glyphs: value references two glyphs; greeting is pinned in local_glyphs,
+        # name is provided only as an example the user should override.
+        _BLOCK_GLYPHS: _make_source_text_block("${greeting} ${name}"),
+        # Example only: no value in configuration_values; example_values shows
+        # what a user might start with.
+        _BLOCK_EXAMPLE: _make_source_text_block(""),
     },
-    local_glyphs={"example_text": "hello world"},
-    example_values={_BASIC_BLOCK_ID: {TEXT: "hello world"}},
-    example_glyphs={"example_text": "hello world"},
+    local_glyphs={"greeting": "hello"},
+    example_values={_BLOCK_EXAMPLE: {TEXT: "text from example values"}},
+    example_glyphs={"name": "world"},
 )
 
 plugin = lambda: Plugin(
--- a/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
+++ b/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
@@ -236,10 +236,21 @@ _testExclusion = BlueprintTemplate(
     },
 )
 
+_BLOCK_REMAP = BlockInstanceId("text_remap")
+
+_testRemapping = BlueprintTemplate(
+    display_name="testRemapping",
+    display_description="A minimal template used to verify glyph-name remapping at install time.",
+    blocks={
+        _BLOCK_REMAP: _make_source_text_block("${pluginGlyphOld}"),
+    },
+    local_glyphs={"localOld": "${pluginGlyphOld}"},
+)
+
 plugin = lambda: Plugin(
     catalogue=catalogue(),
     validator=validator,
     expander=expander,
     compiler=compiler,
-    blueprint_templates=(_testBasic, _testExclusion),
+    blueprint_templates=(_testBasic, _testExclusion, _testRemapping),
 )
--- a/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
+++ b/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
@@ -247,10 +247,24 @@ _testRemapping = BlueprintTemplate(
     local_glyphs={"localOld": "${pluginGlyphOld}"},
 )
 
+_BLOCK_FAIL_VAL = BlockInstanceId("fail_val_block")
+
+_testFailValidation = BlueprintTemplate(
+    display_name="testFailValidation",
+    display_description="A template that references a non-existent factory and always fails validation.",
+    blocks={
+        _BLOCK_FAIL_VAL: BlockInstance(
+            factory_id=PluginBlockFactoryId(plugin=SelfPluginId, factory=BlockFactoryId("nonexistent_factory")),
+            configuration_values={},
+            input_ids={},
+        ),
+    },
+)
+
 plugin = lambda: Plugin(
     catalogue=catalogue(),
     validator=validator,
     expander=expander,
     compiler=compiler,
-    blueprint_templates=(_testBasic, _testExclusion, _testRemapping),
+    blueprint_templates=(_testBasic, _testExclusion, _testRemapping, _testFailValidation),
 )
--- a/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
+++ b/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
@@ -11,10 +11,16 @@ from fiab_core.fable import (
     BlockFactoryCatalogue,
     BlockFactoryId,
     BlockInstance,
+    BlockInstanceId,
     BlockInstanceOutput,
+    BlueprintTemplate,
     ConfigurationOptionId,
     ConfigurationOptionRestriction,
     NoOutput,
+    PluginBlockFactoryId,
+    PluginCompositeId,
+    PluginId,
+    PluginStoreId,
     RawOutput,
 )
 from fiab_core.plugin import BlockValidation, Error, Plugin
@@ -190,3 +196,34 @@ def compiler(lookup: ActionLookup, instance: BlockInstance) -> Either[Action, Er
 
 
 plugin = lambda: Plugin(catalogue=catalogue(), validator=validator, expander=expander, compiler=compiler)
+
+
+_SELF_PLUGIN_ID = PluginCompositeId(store=PluginStoreId("__self__"), local=PluginId("__self__"))
+
+_BASIC_BLOCK_ID = BlockInstanceId("text_source")
+
+_testBasic = BlueprintTemplate(
+    display_name="testBasic",
+    display_description="A minimal test template with a single source_text block.",
+    blocks={
+        _BASIC_BLOCK_ID: BlockInstance(
+            factory_id=PluginBlockFactoryId(
+                plugin=_SELF_PLUGIN_ID,
+                factory=BlockFactoryId("source_text"),
+            ),
+            configuration_values={TEXT: "{{ example_text }}"},
+            input_ids={},
+        ),
+    },
+    local_glyphs={"example_text": "hello world"},
+    example_values={_BASIC_BLOCK_ID: {TEXT: "hello world"}},
+    example_glyphs={"example_text": "hello world"},
+)
+
+plugin = lambda: Plugin(
+    catalogue=catalogue(),
+    validator=validator,
+    expander=expander,
+    compiler=compiler,
+    blueprint_templates=(_testBasic,),
+)
--- a/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
+++ b/backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
@@ -226,10 +226,20 @@ _testBasic = BlueprintTemplate(
     example_glyphs={"name": "world"},
 )
 
+_BLOCK_EXCL = BlockInstanceId("text_excl")
+
+_testExclusion = BlueprintTemplate(
+    display_name="testExclusion",
+    display_description="A minimal template used to verify admin exclusion via the settings route.",
+    blocks={
+        _BLOCK_EXCL: _make_source_text_block("exclusion test"),
+    },
+)
+
 plugin = lambda: Plugin(
     catalogue=catalogue(),
     validator=validator,
     expander=expander,
     compiler=compiler,
-    blueprint_templates=(_testBasic,),
+    blueprint_templates=(_testBasic, _testExclusion),
 )
--- a/backend/src/forecastbox/domain/blueprint/service.py
+++ b/backend/src/forecastbox/domain/blueprint/service.py
@@ -51,7 +51,7 @@ from forecastbox.domain.blueprint.types import BlueprintId
 from forecastbox.domain.glyphs import global_db, resolution
 from forecastbox.domain.glyphs.exceptions import GlyphCircularReferenceError
 from forecastbox.domain.glyphs.intrinsic import get_values_and_examples
-from forecastbox.domain.glyphs.resolution import ExtractedGlyphs, expand_glyph_values, merge_glyph_values
+from forecastbox.domain.glyphs.resolution import ExtractedGlyphs, expand_glyph_values, merge_glyph_values, remap_glyph_names
 from forecastbox.domain.plugin.manager import PluginManager
 from forecastbox.utility.auth import AuthContext
 from forecastbox.utility.graph import topological_order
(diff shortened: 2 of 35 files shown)


backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
```diff
@@ -190,3 +196,34 @@ def compiler(lookup: ActionLookup, instance: BlockInstance) -> Either[Action, Er
 
 
 plugin = lambda: Plugin(catalogue=catalogue(), validator=validator, expander=expander, compiler=compiler)
+
+
+_SELF_PLUGIN_ID = PluginCompositeId(store=PluginStoreId("__self__"), local=PluginId("__self__"))
```
tmi: I quite like this. Every plugin author will have use to something like this -- lets then put it to fiab-core, right after the BlueprintTemplate class, name it SelfPluginId, and add a doc comment to it "use this when building the blueprint templates"

backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
```diff
@@ -190,3 +196,34 @@ def compiler(lookup: ActionLookup, instance: BlockInstance) -> Either[Action, Er
...
+_testBasic = BlueprintTemplate(
+    display_name="testBasic",
+    display_description="A minimal test template with a single source_text block.",
+    blocks={
+        _BASIC_BLOCK_ID: BlockInstance(
+            factory_id=PluginBlockFactoryId(
+                plugin=_SELF_PLUGIN_ID,
+                factory=BlockFactoryId("source_text"),
+            ),
+            configuration_values={TEXT: "{{ example_text }}"},
```
tmi: are you sure this is the right syntax for glyphs? While we use jinja, I believe we use the $-syntax, check it out

backend/packages/fiab-plugin-test/src/fiab_plugin_test/__init__.py
```diff
@@ -190,3 +196,34 @@ def compiler(lookup: ActionLookup, instance: BlockInstance) -> Either[Action, Er
...
+            input_ids={},
+        ),
+    },
+    local_glyphs={"example_text": "hello world"},
+    example_values={_BASIC_BLOCK_ID: {TEXT: "hello world"}},
+    example_glyphs={"example_text": "hello world"},
```
tmi: this mixes up too many things in a conflicting manner. What is the final value of the TEXT configuration value? Does it come from the configuration_values in blocks, or from the value in `local_glyphs`, or from the example_values, or from the example glyphs? This is confusing, we need to make it clear.

use three source_text blocks, one with fixed configuration option (no glyphs, no examples), another with a configuration option containing two glyphs "${greeting} ${name}" and a value in local glyphs (greeting=hello) and a value in example glyphs (name=world), and last one with only example_values (text="text from example values")

backend/src/forecastbox/domain/plugin/db.py
```diff
@@ -0,0 +1,84 @@
...
+
+Uses the same session maker as ``forecastbox.schemata.jobs`` so that all tables
+share a single SQLite connection pool and in-process tests can monkeypatch
+``async_session_maker`` to inject an in-memory database.
+
+This table records unversioned app state (install history, per-plugin config).
+Writes are idempotent: insert on first install with empty-default columns for
+excluded_templates / glyph_remapping / template_errors; update version /
+updated_at / error on subsequent installs without clobbering the columns owned
+by later tasks (04/05/06).
```
tmi: dont mention task numbers in comments -- task numbers are ephemeral, but this code will live for long

backend/src/forecastbox/domain/plugin/db.py
```diff
@@ -0,0 +1,84 @@
...
+logger = logging.getLogger(__name__)
+
+
+async def upsert_plugin_state(*, plugin_id: str, version: str | None, error: str | None) -> None:
+    """Insert or update the PluginState row for ``plugin_id``.
+
+    On first install: creates a row with empty ``excluded_templates`` / ``glyph_remapping``
+    defaults and ``template_errors=None``.
+    On subsequent installs: updates ``version``, ``updated_at``, and ``error`` only --
+    ``excluded_templates`` and ``glyph_remapping`` are not touched (owned by tasks 04/05).
```
tmi: dont mention task numbers in comments -- task numbers are ephemeral, but this code will live for long

backend/src/forecastbox/domain/plugin/db.py
```diff
@@ -0,0 +1,84 @@
...
+
+import forecastbox.schemata.jobs as _jobs_module
+from forecastbox.schemata.jobs import PluginState
+from forecastbox.utility.db import dbRetry, querySingle
+from forecastbox.utility.time import current_time
+
+logger = logging.getLogger(__name__)
+
+
+async def upsert_plugin_state(*, plugin_id: str, version: str | None, error: str | None) -> None:
```
tmi: version should always be str, never None

backend/src/forecastbox/domain/plugin/manager.py
```diff
@@ -126,14 +150,20 @@ def load_plugins(plugins: PluginsSettings) -> None:
 
 
 def update_single(pluginId: PluginCompositeId, pluginSettings: PluginSettings, install: bool, version: Version | None) -> None:
+    plugin_id_str = PluginCompositeId.to_str(pluginId)
     try:
         if install:
-            install_plugin_compatibly(pluginSettings.pip_source, version)
+            try:
+                install_plugin_compatibly(pluginSettings.pip_source, version)
```
tmi: there is one issue with this -- try_install, which ends up being called by this, does not raise, but only logs error. I want you to revisit the whole install_plugin_compatibly and its means of failure. Ideally, change the return type to Either, and parse the error type. And additionally inspect other callsites of the function and make sure its consistently consumed

backend/src/forecastbox/domain/plugin/manager.py
```diff
@@ -188,6 +217,7 @@ class PluginsStatus(FiabBaseModel):
     plugin_errors: dict[PluginCompositeId, str]
     plugin_versions: dict[PluginCompositeId, str]
     plugin_updatedatetime: dict[PluginCompositeId, str]
+    plugin_install_errors: dict[str, str] = {}
```
tmi: hmm, lets not separate like this: keep just a single plugin_errors, and merge at the time of building this class. The key of the plugin install errors should be PluginCompositeId anyway. If there would be two values for a single key (which is not expected), just concatenate with `;`

backend/src/forecastbox/entrypoint/app.py
```diff
@@ -45,6 +45,8 @@
 
 @asynccontextmanager
 async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
+    import asyncio
```
tmi: import outside of the function body, no reason for this

backend/src/forecastbox/schemata/jobs.py
```diff
@@ -199,6 +199,27 @@ class ExperimentNext(Base):
...
+    Written and updated during plugin install; never versioned.
+
+    Columns owned by later tasks (excluded_templates, glyph_remapping, template_errors)
+    are defined here with safe defaults to avoid schema changes later (no migrations).
+    """
+
+    __tablename__ = "plugin_state"
+
+    plugin_id = Column(String(255), primary_key=True, nullable=False)
+    version = Column(String(255), nullable=True)
```
tmi: name this plugin_version, to separate from db entity versioning

backend/src/forecastbox/schemata/jobs.py
```diff
@@ -199,6 +199,27 @@ class ExperimentNext(Base):
...
 
 
+class PluginState(Base):
+    """Persisted install state for each configured plugin.
+
+    One row per plugin, keyed by the PluginCompositeId rendered as ``store:local``.
+    Written and updated during plugin install; never versioned.
+
+    Columns owned by later tasks (excluded_templates, glyph_remapping, template_errors)
+    are defined here with safe defaults to avoid schema changes later (no migrations).
```
tmi: delete this comment about "later task", this is related to task progression, not to actual code

backend/src/forecastbox/schemata/jobs.py
```diff
@@ -199,6 +199,27 @@ class ExperimentNext(Base):
...
+    """
+
+    __tablename__ = "plugin_state"
+
+    plugin_id = Column(String(255), primary_key=True, nullable=False)
+    version = Column(String(255), nullable=True)
+    updated_at = Column(UTCDateTime, nullable=False)
+    error = Column(String(4096), nullable=True)
```
tmi: this should be named install_error

backend/src/forecastbox/schemata/jobs.py
```diff
@@ -199,6 +199,27 @@ class ExperimentNext(Base):
...
+
+    Columns owned by later tasks (excluded_templates, glyph_remapping, template_errors)
+    are defined here with safe defaults to avoid schema changes later (no migrations).
+    """
+
+    __tablename__ = "plugin_state"
+
+    plugin_id = Column(String(255), primary_key=True, nullable=False)
+    version = Column(String(255), nullable=True)
+    updated_at = Column(UTCDateTime, nullable=False)
```
tmi, line 213: is this field being used when the user calls the route for plugin detail where there is some field like update or install time? Or does the code still use the original path which queries file system ctime? If the latter, refactor to use this

backend/src/forecastbox/domain/plugin/manager.py
```diff
@@ -77,6 +80,15 @@ def load_single(plugin: PluginSettings) -> Plugin | str:
     return "\n".join(errors)
 
 
+def _run_async_from_thread(coro: object) -> object:  # type: ignore[type-arg]
+    """Dispatch *coro* to the event loop stashed on PluginManager and block until done."""
+    loop = PluginManager.loop
+    if loop is None:
+        logger.warning("PluginManager.loop is not set; skipping DB write")
```
tmi: this warrants a crash, not a warning

and add there a " # TODO -- send an event to bring down the whole app, not just the thread in question"

backend/src/forecastbox/utility/packages.py
```diff
@@ -106,19 +106,26 @@ def try_updatedatetime(pip_source: str) -> str:
         return "unknown"
 
 
-def try_install(packages: list[str]) -> None:
-    """Run ``uv pip install`` with the given pkg specs."""
+def try_install(packages: list[str]) -> str | None:
```
tmi: use the Either class here

backend/src/forecastbox/domain/plugin/manager.py
```diff
@@ -102,20 +103,17 @@ def load_plugins(plugins: PluginsSettings) -> None:
...
+                install_error = install_plugin_compatibly(pluginSettings.pip_source, None)
+            else:
+                if try_import(pluginSettings.module_name) is None:
+                    logger.info(f"installing {pluginSettings.module_name} for the first time")
+                    install_error = install_plugin_compatibly(pluginSettings.pip_source, None)
 
             if install_error is not None:
-                _run_async_from_thread(upsert_plugin_state(plugin_id=plugin_id_str, version=None, error=install_error))
+                logger.error(f"install failed for {pluginKey}: {install_error}")
+                _run_async_from_thread(upsert_plugin_state(plugin_id=plugin_id_str, version="unknown", install_error=install_error))
```
tmi: dont ever use `unknown` as the install version. Improve the `try_install` to return the versions of the installed packages -- use https://github.com/ecmwf/earthkit-workflows/blob/main/src/cascade/executor/runner/packages.py#L134 in there

backend/src/forecastbox/domain/plugin/manager.py
```diff
@@ -238,26 +234,36 @@ async def status_full() -> PluginsStatus:
...
+                plugin_id = PluginCompositeId.from_str(state.plugin_id)  # type: ignore[arg-type]
+            except Exception:
+                logger.warning(f"could not parse plugin_id {state.plugin_id!r} from DB; skipping")
+                continue
+            if state.install_error is not None:  # type: ignore[misc]
+                existing = plugin_errors.get(plugin_id)
+                plugin_errors[plugin_id] = (
+                    f"{existing}; {state.install_error}" if existing else state.install_error  # type: ignore[misc]
+                )
+            if state.updated_at is not None:
```
tmi: this should never be None, should it? We set Nullable false on the db, right? This is then redundant, is it?

backend/src/forecastbox/domain/plugin/manager.py
```diff
@@ -238,26 +234,36 @@ async def status_full() -> PluginsStatus:
...
-            plugin_versions = {}
-            plugin_updatedatetime = {}
+            plugin_errors: dict[PluginCompositeId, str] = {}
+            plugin_versions: dict[PluginCompositeId, str] = {}
+            plugin_updatedatetime: dict[PluginCompositeId, str] = {}
         else:
             status = status_brief()
             plugin_errors = dict(PluginManager.errors)
             plugin_versions = dict(PluginManager.versions)
             plugin_updatedatetime = dict(PluginManager.updatedatetime)
```
tmi: you have correctly put usage of the db's updated_at in place -- but left the original file-based pathway in place. I want you to remove the PluginManager's updatedatetime field and all methods that interact with it -- it should *not* be necessary anymore

you can leave the timestamp detecting methods in place

backend/src/forecastbox/domain/blueprint/service.py
```diff
@@ -357,6 +357,44 @@ def template_to_builder(template: BlueprintTemplate, plugin_id: PluginCompositeI
...
+
+    Returns *builder* unchanged (same object) when *mapping* is empty.
+    """
+    if not mapping:
+        return builder
+
+    new_blocks: dict[BlockInstanceId, BlockInstance] = {}
+    for block_id, block in builder.blocks.items():
+        new_config = {opt_id: remap_glyph_names(val, mapping) for opt_id, val in block.configuration_values.items()}
+        new_blocks[block_id] = BlockInstance(
```
tmi: dont create a new instance, use pydantic model copy -- thats more robust towards future changes to BlockInstance fields

backend/src/forecastbox/domain/blueprint/service.py
```diff
@@ -357,6 +357,44 @@ def template_to_builder(template: BlueprintTemplate, plugin_id: PluginCompositeI
...
+            input_ids=dict(block.input_ids),
+        )
+
+    new_local_glyphs: dict[str, str] = {}
+    for key, val in builder.local_glyphs.items():
+        new_key = mapping.get(key, key)
+        new_val = remap_glyph_names(val, mapping)
+        new_local_glyphs[new_key] = new_val
+
+    return BlueprintBuilder(
```
tmi: dont create a new instance, use pydantic model copy -- thats more robust towards future changes to BlueprintBuilder fields

backend/src/forecastbox/domain/plugin/manager.py
```diff
@@ -107,25 +107,24 @@ async def _ingest_plugin_templates(plugin_id: PluginCompositeId, plugin: Plugin)
 
     Excluded templates (per ``PluginState.excluded_templates``) are skipped and
     any existing plugin-owned blueprint row with that ``display_name`` is
-    soft-deleted.  Non-excluded templates are upserted as normal.
+    soft-deleted.  Non-excluded templates have their glyph names rewritten by
+    ``remap_builder_glyphs`` when a non-empty ``glyph_remapping`` is stored for
+    the plugin, then are upserted as normal.
```
tmi, line 124: this introduces another dependency of plugin domain on blueprint domain -- consult the commit d8bcce0d039034a38528fb36daed88a234339faa where we extended the docs accordingly. You need to:
1/ update this place where you add another dependency hierarchy breach with an appropriate note
2/ update the listing of breaches in the inits which the mentioned commit updated

backend/src/forecastbox/domain/blueprint/service.py
```diff
@@ -357,6 +357,42 @@ def template_to_builder(template: BlueprintTemplate, plugin_id: PluginCompositeI
...
+    """Return a copy of ``builder`` with example values/glyphs overlaid for validation only.
+
+    Example configuration values fill in (without overwriting) the per-block
+    ``configuration_values``; example glyphs are merged into ``local_glyphs``.
+    The result is fed to ``validate_expand(validate_only=True)``; it is never persisted.
+
+    Overlay precedence: example values and glyphs only fill keys that are absent in
+    the builder -- they never overwrite an explicit template value.  This matches the
+    intended semantics: example values are defaults that the user is expected to
+    override, so anything already set in the template takes priority.
```
tmi: the precedence should actually go the other way, the examples should win. At this stage, the validation is for original plugin templates, that is, user has not had a chance to override any default value. The default values are in place as "reasonable defaults", but that does not necessarily make them good example candidates, hence we allow for example-overrides-default

backend/src/forecastbox/domain/plugin/manager.py
```diff
@@ -314,6 +343,7 @@ async def status_full() -> PluginsStatus:
             plugin_errors = dict(PluginManager.errors)
     plugin_versions: dict[PluginCompositeId, str] = {}
     plugin_updatedatetime: dict[PluginCompositeId, str] = {}
+    plugin_template_errors: dict[PluginCompositeId, dict[str, str]] = {}
```
tmi: we dont want standalone template errors -- merge it to existing plugin errors during handling of the request, as it is already a merge of two sources

PR #529, opened 2026-07-02
Add GET /plugin/templateExampleValues route

tmi:
Returns example_values and example_glyphs for a blueprint template from a loaded plugin, with glyph remapping applied (matching the remapping done at template ingest time).

- 404 when plugin is not loaded or display_name not found in the plugin
- 403 when the template is excluded by admin settings
- remapping applied to example_values (values via remap_glyph_names) and example_glyphs (keys renamed via mapping, values via remap_glyph_names)

Uses remap_glyph_names from domain/glyphs; notes the new domain dependency in both domain/__init__.py files.

Unit tests cover happy path, remapping, and all error cases. Integration tests verify real examples, the 403 on excluded template, and 404 for unknown display name and unknown plugin.

### Contributor Declaration

### DIFF
--- a/backend/src/forecastbox/domain/glyphs/__init__.py
+++ b/backend/src/forecastbox/domain/glyphs/__init__.py
@@ -4,7 +4,5 @@ user can use the Key across all their Blueprints, and these are then replaced
 by the Value at runtime when executing a Run. This facilitates value reusal.
 
 Depends on no other domain.
-Depended on by Blueprint, Experiment, and Run (each of them needs to resolve Glyphs).
-Also depended on by the Plugin route (``routes/plugins.py``) for ``remap_glyph_names``,
-used when serving template example values via ``/plugin/templateExampleValues``.
+Depended on by Blueprint, Experiment, and Run (each of them needs to resolve Glyphs), and Plugin (for glyph remapping).
 """
--- a/backend/src/forecastbox/domain/glyphs/__init__.py
+++ b/backend/src/forecastbox/domain/glyphs/__init__.py
@@ -5,4 +5,6 @@ by the Value at runtime when executing a Run. This facilitates value reusal.
 
 Depends on no other domain.
 Depended on by Blueprint, Experiment, and Run (each of them needs to resolve Glyphs).
+Also depended on by the Plugin route (``routes/plugins.py``) for ``remap_glyph_names``,
+used when serving template example values via ``/plugin/templateExampleValues``.
 """
--- a/backend/src/forecastbox/domain/plugin/__init__.py
+++ b/backend/src/forecastbox/domain/plugin/__init__.py
@@ -5,12 +5,8 @@ building and validation, and Run execution.
 Now also owns persisted install state: each configured plugin's install outcome (version, timestamp,
 install error) is recorded in the ``plugin_state`` DB table via ``domain/plugin/db.py``.
 
-Depends on no other domain.
+Depends on Glyph domain (for glyph remapping).
 Depended on by Blueprint and Run domains -- they need to extract individual plugin validations and compilers, and to validate fiab-core version.
 
 Note: there is a dependency circularity where this domain *depends on* Blueprint, for validating imported blueprint templates and for remapping glyph names in those templates. This will be fixed later by refactoring into events.
-
-Note: the plugin route (``routes/plugins.py``) depends on the Glyph domain (``domain/glyphs``) for
-``remap_glyph_names``, used to apply glyph remapping to template example values and glyphs in the
-``/plugin/templateExampleValues`` endpoint.
 """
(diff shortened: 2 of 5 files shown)


backend/src/forecastbox/domain/glyphs/__init__.py
```diff
@@ -5,4 +5,6 @@
 
 Depends on no other domain.
 Depended on by Blueprint, Experiment, and Run (each of them needs to resolve Glyphs).
+Also depended on by the Plugin route (``routes/plugins.py``) for ``remap_glyph_names``,
+used when serving template example values via ``/plugin/templateExampleValues``.
```
tmi: make it more brief and merge with the line above -- like `", and Plugin (for glyph remapping)"`.

backend/src/forecastbox/domain/plugin/__init__.py
```diff
@@ -9,4 +9,8 @@
 Depended on by Blueprint and Run domains -- they need to extract individual plugin validations and compilers, and to validate fiab-core version.
 
 Note: there is a dependency circularity where this domain *depends on* Blueprint, for validating imported blueprint templates and for remapping glyph names in those templates. This will be fixed later by refactoring into events.
+
+Note: the plugin route (``routes/plugins.py``) depends on the Glyph domain (``domain/glyphs``) for
+``remap_glyph_names``, used to apply glyph remapping to template example values and glyphs in the
+``/plugin/templateExampleValues`` endpoint.
```
tmi: make it more brief and merge with the "Depended on by " line, like ", and Plugin domain (for glyph remapping)".

backend/src/forecastbox/routes/plugins.py
```diff
@@ -226,3 +234,55 @@ async def update_plugin_settings_endpoint(
     if result:
         raise HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=result)
     return get_catalogue_redirect(request)
+
+
+class TemplateExampleValuesResponse(FiabBaseModel):
+    example_values: dict[str, dict[str, str]]
+    """Per-block example configuration values, keyed by block instance id then option id."""
+    example_glyphs: dict[str, str]
+    """Example glyph name-to-value pairs the user is expected to override."""
```
tmi, line 243: can we go not with str keys, but use the narrow types, like ConfigurationOptionId and BlockInstaceId, where they exist? They are NewType constructs which are internally string, meaning serde should be fine for them

backend/src/forecastbox/routes/plugins.py
```diff
@@ -226,3 +234,55 @@ async def update_plugin_settings_endpoint(
...
+        raise HTTPException(status_code=404, detail=f"Plugin {PluginCompositeId.to_str(pluginCompositeId)!r} not loaded")
+
+    template = next((t for t in plugin.blueprint_templates if t.display_name == displayName), None)
+    if template is None:
+        raise HTTPException(
+            status_code=404, detail=f"Template {displayName!r} not found in plugin {PluginCompositeId.to_str(pluginCompositeId)!r}"
+        )
+
+    plugin_id_str = PluginCompositeId.to_str(pluginCompositeId)
+    plugin_state = await get_plugin_state(plugin_id_str)
```
tmi, line 270: can this return None? Can this raise? Guard for it, return 404 again, "plugin not installed"

backend/tests/integration/test_blueprint.py
```diff
@@ -316,6 +337,27 @@ def verify_ok(data: dict) -> dict | None:
     assert "pluginGlyphNew" in local_glyphs.get("localNew", ""), f"Expected localNew value to reference pluginGlyphNew, got: {local_glyphs}"
 
 
+def test_template_example_values_not_found(backend_client_with_auth: httpx.Client) -> None:
+    """Non-existent display name and non-existent plugin return 404 from templateExampleValues."""
```
tmi: merge this with the `test_plugin_template_exclusion` above, we want few tests with longer scope in the integration tests

tmi:
cc @liefra -- this adds the new route for providing example values and glyphs for a given blueprint template

its a bare minimum so that you can start experimenting with UI. We can change the interface if you would prefer for any reason -- now to have the final example, you need to get the blueprint, get the example values, and merge yourself. We can do it other way, that this endpoint would return the builder with the examples is place. But I think you would want to distinguish, visually, between which value is a default and which is just example, hence the existing response type would not fit anyway. Nevertheless, happy to change in any direction :)

note that the example values and glyphs are not persisted, we read them from the plugin wheel every time, and apply remapping/exclusion on the fly. This implies that any change to example values (eg, "defining your own template with examples") by the client are not possible. If that were to become a need, I would probably promote it to its own domain entity, with its own router etc

liefra:
Thanks for letting me know. Will try it out soon and get back then

PR #531, opened 2026-07-02
Feat/geo types

tmi:
### Contributor Declaration

### DIFF
--- a/backend/packages/fiab-core/src/fiab_core/types.py
+++ b/backend/packages/fiab-core/src/fiab_core/types.py
@@ -71,7 +71,7 @@ class FableType(ABC):
         """
         type_expr = type_expr.lstrip()
 
-        # Atomic types: startswith + word-boundary check (next char not alnum or '_')
+        # Atomic types: "datetime" must precede "date" to avoid prefix ambiguity.
         _ATOMIC = [
             ("datetime", DatetimeType),
             ("date", DateType),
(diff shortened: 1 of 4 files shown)


backend/packages/fiab-core/src/fiab_core/types.py
```diff
@@ -50,57 +53,127 @@ def serialize(self) -> str:
...
+            ("date", DateType),
+            ("float", FloatType),
+            ("int", IntType),
+            ("str", StringType),
+            ("country", CountryType),
+            ("bbox", BoundingBoxType),
+        ]
+        for name, factory in _ATOMIC:
+            n = len(name)
+            if type_expr.startswith(name) and (len(type_expr) == n or not (type_expr[n].isalnum() or type_expr[n] == "_")):
```
tmi: why are there the isalnum() or `"_"`? I mean, `str42` or `int_` is a wrong type, but we can simply decide at this stage that the "str"/"int" part is fine, and leave the "42"/"_" as remainder, which will most likely cause a parsing crash

lets keep this simple, and only check for the startswith here -- or do I overlook some cornercase?

backend/packages/fiab-core/src/fiab_core/types.py
```diff
@@ -50,57 +53,127 @@ def serialize(self) -> str:
...
+                return (factory(), type_expr[n:])
+
+        # enumClosed[...]
+        if type_expr.startswith("enumClosed["):
+            bs = 10  # index of '['
+            pos = _find_matching_bracket(type_expr[bs:])
+            if pos == -1:
+                raise NotFableType("Unmatched '[' in enumClosed expression")
+            inner = type_expr[bs + 1 : bs + pos]
+            remainder = type_expr[bs + pos + 1 :]
```
tmi: the hardcoded `bs` and the `-1` check and explicit parsing looks too heavy handed. Lets change as follows:

rename `_find_matching_bracket` to `_split_by_brackets` -- that function will accept a string like "prefix[...]...]", and output prefix, inner, remainder. If it does not find the opening or the matching closing bracket, it will raise NotFableType(f"Unmatched brackets in {prefix} expression") etc

then all usage in enumClosed, enumOpen, list, union, will look like "_, inner, remainder = _split_by_brackets(type_expr)". And actually, this would allow you to unify the enumClosed and enumOpen completely: you'd have just `_ENUMS = {"enumClosed": ClosedEnumType, "enumOpen": OpenEnumType}`, and make the code like `if any(type_expr.startswith(e for e in _ENUMS.keys()): _split_by_brackets, ..., return _ENUMS[prefix](items)`

backend/packages/fiab-core/src/fiab_core/types.py
```diff
@@ -50,57 +53,127 @@ def serialize(self) -> str:
...
+            pos = _find_matching_bracket(type_expr[bs:])
+            if pos == -1:
+                raise NotFableType("Unmatched '[' in list expression")
+            inner = type_expr[bs + 1 : bs + pos]
+            remainder = type_expr[bs + pos + 1 :]
+            inner_type, inner_remainder = FableType.parse(inner)
+            if inner_remainder.strip():
+                raise NotFableType(f"Unexpected content after inner type in list: {inner_remainder!r}")
+            if isinstance(inner_type, (ListType, UnionType)):
+                raise NotFableType("Nested lists and union types inside list are not supported")
```
tmi: is this exception raising still meaningful? I mean, the whole purpose of the recent prompt was to enable this nesting -- am I missing some reason for why it won't work?

backend/packages/fiab-core/src/fiab_core/types.py
```diff
@@ -50,57 +53,127 @@ def serialize(self) -> str:
...
+            first = True
+            while remaining:
+                remaining = remaining.lstrip()
+                if not first:
+                    if not remaining.startswith(","):
+                        raise NotFableType(f"Expected ',' between union member types, got {remaining!r}")
+                    remaining = remaining[1:].lstrip()
+                first = False
+                t, remaining = FableType.parse(remaining)
+                remaining = remaining.rstrip()
```
tmi: why do we rstrip here? Isn't it enough to call lstrip as we already do? The remaining does not change from the right, hence all invocations except the first one can't do anything, and the first one can be handled instead by the _last_ lstrip call, no?

and it would actually be better to call the `remaining.lstrip()` here than at the beginning of the while loop, because if the remaining is only whitespaces, we dont want to enter the loop anymore

backend/packages/fiab-core/src/fiab_core/types.py
```diff
@@ -50,57 +53,127 @@ def serialize(self) -> str:
...
+            bs = 5  # index of '['
+            pos = _find_matching_bracket(type_expr[bs:])
+            if pos == -1:
+                raise NotFableType("Unmatched '[' in union expression")
+            inner = type_expr[bs + 1 : bs + pos].strip()
+            remainder = type_expr[bs + pos + 1 :]
+            if not inner:
+                raise NotFableType("union must contain at least one type")
+            member_types: list[FableType] = []
+            remaining = inner
```
tmi, line 111: you should call lstrip here as well, we dont want to enter the loop with a string containing just spaces

or even better, you can call strip on all strings returned by the _separate_by_brackets, that would simplify all methods that utilize it

PR #536, opened 2026-07-07
[plugin] Initial set of product blocks

jinmannwong:
### Description
- Add product blocks for computating ensemble mean and standard deviation, predefined threshold probabilities (those with param ids), custom threshold probabilities and thermal indices
- Changes labelling in `PARAM` dimension to use param id's instead of shortname, and displays a combination of both, in the format "{shortname}-{paramid}" in the UI

### Contributor Declaration

#580


backend/packages/fiab-plugin-ecmwf/pyproject.toml
```diff
@@ -20,6 +20,8 @@ dependencies = [
     "fiab-core>=1,<2", # auto-updateable -- do not remove this comment
     "earthkit-workflows>=0.9.0",
     "earthkit-workflows-anemoi>=0.7.2",
+    "pproc-core @ git+https://git@github.com/ecmwf/pproc.git@feature/mono-repo#subdirectory=pproc-core",
+    "pymetkit @ git+https://git@github.com/ecmwf/metkit.git@feature/paramdb-api",
```
tmi, line 24: is there a plan to publish both of those to PyPI? Happy to go with this as-is for a while since we have a lock which pins the concrete commit hash -- but given that these are branches, it may eventually happen that after you merge the PR and delete the branch the commit gets deleted, leading to uninstallable project... so the sooner we switch to PyPI installs, the better

jinmannwong: Yes, I agree. For pproc-core, I am hoping we can publish something soon. I also ask Adam about the pymetkit branch and he said he is hoping to get it merged soon as well.

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/products/blocks.py
```diff
@@ -0,0 +1,340 @@
...
+    _param_id_to_param_key,
+    _param_key_to_param_id,
+)
+from fiab_plugin_ecmwf.qubed_utils import axes, collapse, contains, coxpand, datacubes, from_datacubes, select
+
+logger = logging.getLogger(__name__)
+
+
+def load_pproc_schema(cache_size: int) -> Schema:
+    with as_file(files("fiab_plugin_ecmwf.products.pproc") / "schema.yaml") as pproc_schema:
```
tmi: minor improvement suggestion: replace this with
```
import fiab_plugin_ecmwf.products.pproc
[...]
  files(fiab_plugin_ecmwf.products.pproc)
```
that would make it easier for typechecker to warn when a refactoring would have moved it; `files` can take the imported module object too

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/templates/ifs_ensemble_statistics.py
```diff
@@ -72,7 +72,6 @@
             factory_id=BlockFactoryId("ensembleStatistics"),
             instance=BlockInstance(
                 configuration_values={
-                    ConfigurationOptionId("param"): "2t",
```
tmi, line 75: is this intended? If yes we should update the docstring, and move it up to example glyphs, no?

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/block_utils.py
```diff
@@ -0,0 +1,87 @@
...
+SPLITBY = ConfigurationOptionId("splitby")
+STATISTIC = ConfigurationOptionId("statistic")
+STEP = ConfigurationOptionId("step")
+THRESHOLD = ConfigurationOptionId("threshold")
+TIME = ConfigurationOptionId("time")
+TYPE = ConfigurationOptionId("type")
+VALUES = ConfigurationOptionId("values")
+
+
+@functools.cache
```
tmi: why are we caching this like this? Is it because the ParamDB is expensive to construct, or that the lookups take long time? If the former, I'd suggest converting ParamDB to a singleton

I have generally very low view of `functools.cache`, it is often a source of unnecessarily memory consumption and slowdown

in this case it looks really odd -- what we are _additionally_ caching is the `int(param_id)` for every string, and this path has nothing to do with ParamDB

also, what is the plausible set of integers this would be executed with? If small, like <1024, we could simply calculate it at build time and do no business with ParamDB whatsoever :)

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/block_utils.py
```diff
@@ -0,0 +1,87 @@
...
+    try:
+        param_id_int = int(param_id)
+    except ValueError:
+        return param_id
+    db = ParamDB()
+    shortname = db.param_id_to_shortname(param_id_int)
+    return f"{shortname}-{param_id}"
+
+
+@functools.cache
```
tmi: similar here -- this looks like it should be a reasonably fast function. How often are we calling it?

tmi: did a quick microbenchmark of four options, this one is the fastest:
```
def trim_find(s):
    idx = s.find('-')
    return s[idx + 1:] if idx != -1 else s
```

it can do ~1M iterations under 0.1 seconds (for both match/unmatch case, where the unmatch is ~1.6x faster), I hope we are not running that much?

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
```diff
@@ -155,8 +125,9 @@ def compile(
         forecast = block.config_as_str(FORECAST)
         fc_preset = FORECAST_DATASETS[forecast]
         fc_qube = fc_preset.as_qube(ens_dim=ENSEMBLE)
+        paramdb = ParamDB()
```
tmi: bit puzzled, I thought ParamDB is the heavy object to construct because of the caching above -- but apparently not then, right? But if we made a singleton (which may be a good idea anyway to have a less gc), then we could re-use that here

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/templates/ifs_ensemble_statistics.py
jinmannwong: Yes, I removed the `param` configuration option in the EnsembleStatistics block, as it was added before we had the Select block based on parameter. The docstring is still correct as we have an Select block in this template that selects 2t.

backend/uv.lock
```diff
@@ -3170,6 +3151,25 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/e9/b7/5d7ffad4dd6fba63641d9d3f8289e83009d621ff1f487ba923502ffff20f/polytope_client-0.7.9-py3-none-any.whl", hash = "sha256:5531670e3f81b58caef7ffda88082380e81b1cbd78c0a1b60b80899bf746a6d9", size = 52422, upload-time = "2025-10-22T17:17:35.04Z" },
 ]
 
+[[package]]
+name = "pproc-core"
```
tmi, line 3191: its a private repo, right? Or do I just have a bad url? If its private, do we plan to make it public (since up until now all fiab & deps is public). I mean, it exists as a source dist on pypi, and is apache licensed, so it probably should :)

backend/uv.lock
```diff
@@ -3385,30 +3385,17 @@ wheels = [
 
 [[package]]
 name = "pyfdb"
-version = "5.22.0.26"
+version = "0.1.3"
```
tmi: is this correct? 0.1.3 looks like a very odd version. Which drops fdb5lib, adds cffi, ... not expected

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/block_utils.py
tmi: ok, I get that the param ids are those in the artifacts.json or product schemata, so mostly we are talking about the range 100-200... Maybe the best would be to have it as a post-init field on the objects themselves, thats cleaner than caching. In the sense of having the field as `int|str`, where the `int` is the result of `int()` call and the `str` of the ParamDB lookup, instead of holding the `"str"` repr of an int and caching the conversion... thats really odd :)

jinmannwong: This function is called every time we call validate on a block which has restrictions based on parameter, and I thought I saw a minor improvement when caching for the repeated lookups. However, I don't mind removing it for now and we can come back to this if we later find validation is too slow. Your option of calculating it at build time might be the better solution, but I am not sure how large the space of param ids is, so I would need to find that out first.

tmi (approved):
generally lgtm, just a few nits -- can be merged as is, and some of the nits converted to optional TODOs. We could delay solving the perf related ones to when we have the perf benchmarking suite in place (quite high on my list)

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/block_utils.py
jinmannwong: Thanks, I will change it to this.

backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/templates/ifs_ensemble_statistics.py
tmi: ah I see, makes perfect sense, thx!

backend/uv.lock
jinmannwong: pproc-core is not in it's own repo, it will be in the pproc mono repo, but the PR to change pproc to a mono repo has not yet been merged

backend/uv.lock
jinmannwong: Thanks for spotting this, I think we are missing a version of pyfdb/fdb5libs compatible with the latest metkit release.

tmi: that is plausible -- "latest" is generally unreliable term since we are using non-identical bundles. So one needs to find one .D such that there is both fdb x1.y1.z1.D and metkit x2.y2.z2.D released for that (uv lock -P, when given the leaf-libs, should be able to resolve the .D for you)

PR #546, opened 2026-07-10
fix: reorder drawn bbox to earthkit's [W, E, S, N] in map_plot

liefra:
The refactor in PR #543 dropped the reorder that parse_geodomain used to do, so a drawn bounding box reached add_map in wire order [W, S, E, N] but earthkit requires [W, E, S, N]  resulting in cropping the wrong region. 

Restore the reorder for the four-int bbox case, named regions and the `auto` are unaffected.

### Contributor Declaration


tmi:
solved instead in https://github.com/ecmwf/forecast-in-a-box/pull/548

thanks again for finding the bug in the first place! :rocket:

PR #547, opened 2026-07-10
Refactor plugin routes: merge status_full into /plugin/list detail

tmi:
- Add PluginManagerBusy exception to domain/plugin/exceptions.py
- Create domain/plugin/detail.py with the PluginDetail class hierarchy (PluginGenericData, PluginInstallSettings, PluginInstallData, PluginDetail, PluginListing) and the build_plugin_listing() handler.  These classes are both frontend-exposed and used internally; see module docstring.
- Remove PluginsStatus and status_full() from manager.py (logic subsumed by detail.py).  status_brief() and everything else in manager.py is untouched.
- Remove the /plugin/status route and the old /plugin/details route from routes/plugins.py; add GET /plugin/list backed by build_plugin_listing(), returning 503 on PluginManagerBusy.
- Update integration tests to poll GET /status (plugins field from status_brief()) for readiness, and use /plugin/list for plugin-specific assertions (install_data, settings_data, load_errors).
- Relocate status_full unit tests from test_plugin_manager.py to new test_plugin_detail.py covering build_plugin_listing(); replace test_plugin_manager.py content with tests for status_brief/plugins_ready.

### Contributor Declaration

### DIFF
--- a/backend/src/forecastbox/domain/plugin/detail.py
+++ b/backend/src/forecastbox/domain/plugin/detail.py
@@ -1,0 +1,210 @@
+# (C) Copyright 2024- ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Plugin listing detail -- response classes and handler for the GET /plugin/list route.
+
+These classes are both frontend-exposed (serialised as JSON in HTTP responses) and
+consumed internally by the service layer.  Changes to field names or structure affect
+the frontend contract; coordinate with frontend consumers before making breaking changes.
+
+The public entry-point is ``build_plugin_listing()``.  It acquires the PluginManager
+lock for the duration of the in-memory snapshot capture *and* the database reads, so
+that both are taken from the same logical point in time.  This is not a fully
+transactional read (the SQLite reads are not inside a DB transaction), but it is
+sufficient for a consistent UI snapshot.  The lock is held only for the duration of
+these fast operations; long-running work must not be done under it.
+"""
+
+import logging
+
+from fiab_core.fable import PluginCompositeId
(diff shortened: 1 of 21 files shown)


backend/src/forecastbox/domain/plugin/detail.py
```diff
@@ -0,0 +1,210 @@
...
+    return any(e.severity in ("error", "critical") for e in install_errors)
+
+
+def _build_detail(
+    plugin_id: PluginCompositeId,
+    store_entry: PluginStoreEntry | None,
+    remote_info: PluginRemoteInfo | None,
+    plugin_in_memory: Plugin | None,
+    in_memory_errors: PluginErrors,
+    db_state: object,  # PluginState ORM row or None
```
tmi: why dont you use PluginState|None here, instead of `object`? Not that it would improve the typing down there too much, but still would read better than the comment

backend/src/forecastbox/domain/plugin/detail.py
```diff
@@ -0,0 +1,210 @@
...
+) -> PluginDetail:
+    generic_data = PluginGenericData(store_info=store_entry, remote_info=remote_info)
+
+    if db_state is None:
+        return PluginDetail(
+            generic_data=generic_data,
+            load_errors=PluginErrors(list(in_memory_errors)),
```
tmi: this is odd -- if there is no db_state, it means the install likely didnt succeed, meaning there should be no load_errors. Check that in_memory_errors is empty and log a warning if they are not, and either way just give an empty list to in_memory_errors (or even better, introduce a default_factory PluginErrors([]) to the PluginDetail

backend/src/forecastbox/domain/plugin/detail.py
```diff
@@ -0,0 +1,210 @@
...
+        )
+
+    # install_data
+    install_errors = PluginErrors(
+        [PluginError(**e) for e in (db_state.plugin_errors or [])]  # type: ignore[union-attr]
+    )
```
tmi: I think you should be careful here, and separate the db_state.plugin_errors based on source whether they should go to install_errors or load_errors. Even if it currently is that all persisted errors would be install_errors, the structure allows it to change in the future, hence we should do this robustly

PR #559, opened 2026-07-16
Chore/response fields

tmi:
Increased Consistency & Minor Security Gap Fix

1/ Adds created_at, updated_at, user, to the blueprint get/list and experiment get/list to where they were missing
2/ Utilizes the same response schema for blueprint get/list, with the only difference of blueprint list not having the builder filled in for response size reasons
3/ Fixes a minor security gap -- get blueprint did not apply auth context
4/ Frontend adaptation to the field renames -- `created_by` was renamed by the backend to `user` for consistency. Names inside frontend were kept as they were, its just the accesor name change

cc @liefra


liefra:
Thanks. Frontend changes are good to merge

PR #571, opened 2026-07-20
Bug/improve dataset availability

tmi:
Builds on a previous PR which had backend rely on cached values for outputs of completed jobs

In this one, when a gateway has changed since a job has completed (eg due to a laptop restart), we dont even include the other outputs in the available_task_ids,  we instead put them to a new field called lost_task_ids, which is a dict[TaskId, str], where the value is "reason" -- we include like "Gateway Process has changed". In the future, we plan to add like "Gateway has evicted this dataset due to memory pressure", or "Data are not found on the url anymore" (for when the underlying data would be removed -- though this one is dubious bcs of the data-metadata problem)

### DIFF
--- a/backend/src/forecastbox/domain/gateway/service.py
+++ b/backend/src/forecastbox/domain/gateway/service.py
@@ -203,6 +203,28 @@ def status_gateway() -> str:
         assert_never(gateway_connection)
 
 
+def get_current_cascade_proc() -> int | str:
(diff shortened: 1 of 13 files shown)


tmi:
cc @liefra -- just a new field to the RunDetail, you can eventually utilize (though no pressure, this is of very low importance)

PR #600, opened 2026-07-22
Feat/cr rework/ph1

tmi:
### Contributor Declaration

### DIFF
--- a/backend/src/forecastbox/utility/concurrency/manager.py
+++ b/backend/src/forecastbox/utility/concurrency/manager.py
@@ -14,20 +14,29 @@ import threading
 import time
 import traceback
 from collections import deque
-from collections.abc import Callable, Mapping
+from collections.abc import Callable
 from concurrent.futures import Future, ThreadPoolExecutor
 from dataclasses import dataclass
 from enum import StrEnum
-from types import MappingProxyType
-from typing import Any, NewType, TypeVar, cast
+from typing import Any, NewType, Protocol, TypeVar, cast
+
+from pydantic import BaseModel, ConfigDict, SerializeAsAny
 
 from forecastbox.utility.config import ConcurrentPools, ConcurrentThreads, config
+from forecastbox.utility.pydantic import FiabBaseModel
+from forecastbox.utility.structural import freeze_mapping
 
 T = TypeVar("T")
 TaskName = NewType("TaskName", str)
 SyncTask = Callable[[], T]
 ThreadEntrypoint = Callable[[threading.Event], None]
-StatusProvider = Callable[[], object]
+
+
+class ComponentStatusProtocol(Protocol):
+    def is_ready(self) -> bool: ...
+
+
+StatusProvider = Callable[[], ComponentStatusProtocol]
 StopRequest = Callable[[float], None]
 
 
@@ -39,19 +48,15 @@ class ExecutionLifecycle(StrEnum):
     stopped = "stopped"
 
 
-Lifecycle = ExecutionLifecycle
-
-
 @dataclass(frozen=True, eq=True, slots=True)
 class RestartPolicy:
     """A finite restart budget for one managed long-lived thread."""
 
-    def __init__(self, max_restarts: int = 0) -> None:
-        if max_restarts < 0:
-            raise ValueError("max_restarts must be non-negative")
-        object.__setattr__(self, "max_restarts", max_restarts)
-
     max_restarts: int = 0
+
+    def __post_init__(self) -> None:
+        if self.max_restarts < 0:
+            raise ValueError("max_restarts must be non-negative")
 
 
 NEVER_RESTART = RestartPolicy()
@@ -77,24 +82,28 @@ class SubmissionRejected(ExecutionManagerError):
     """Raised when a task cannot be accepted by a managed pool."""
 
 
-@dataclass(frozen=True, eq=True, slots=True)
-class ComponentStatus:
+class StatusModel(FiabBaseModel):
+    model_config = ConfigDict(frozen=True)
+
+
+class ComponentStatus(StatusModel):
     ready: bool
     health: str = "unknown"
     detail: str = ""
 
+    def is_ready(self) -> bool:
+        return self.ready
 
-@dataclass(frozen=True, eq=True, slots=True)
-class MonitoredFailure:
-    pool_name: ConcurrentPools | str
+
+class MonitoredFailure(StatusModel):
+    pool_name: ConcurrentPools
     task_name: TaskName
     exception_type: str
     message: str
     traceback: str
 
 
-@dataclass(frozen=True, eq=True, slots=True)
-class PoolStatus:
+class PoolStatus(StatusModel):
     pool_name: ConcurrentPools
     stage: int
     max_workers: int
@@ -110,13 +119,12 @@ class PoolStatus:
     accepting: bool
 
 
-@dataclass(frozen=True, eq=True, slots=True)
-class ThreadStatus:
-    thread_name: str
+class ThreadStatus(StatusModel):
+    thread_name: ConcurrentThreads
     stage: int
     alive: bool
     ready: bool
-    component_status: object
+    component_status: SerializeAsAny[BaseModel]
     started_at: float | None
     stopped_at: float | None
     uncaught_exception: str | None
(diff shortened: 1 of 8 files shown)


backend/src/forecastbox/utility/config.py
```diff
@@ -341,6 +398,8 @@ class FIABConfig(BaseSettings):
     external: ExternalServicesSettings = Field(default_factory=ExternalServicesSettings)
     backend: BackendSettings = Field(default_factory=BackendSettings)
     cascade: CascadeSettings = Field(default_factory=CascadeSettings)
+    concurrency: ConcurrencySettings = Field(default_factory=ConcurrencySettings)
+    dispatcher: DispatcherSettings = Field(default_factory=DispatcherSettings)
```
tmi: lets actually nest these under BackendSettings

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+from types import MappingProxyType
+from typing import Any, NewType, TypeVar, cast
+
+from forecastbox.utility.config import ConcurrentPools, ConcurrentThreads, config
+
+T = TypeVar("T")
+TaskName = NewType("TaskName", str)
+SyncTask = Callable[[], T]
+ThreadEntrypoint = Callable[[threading.Event], None]
+StatusProvider = Callable[[], object]
```
tmi: `object` is too open ended return type. Since status will end up in http route responses, lets instead put pydantic base model here

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+
+class ExecutionLifecycle(StrEnum):
+    new = "new"
+    starting = "starting"
+    running = "running"
+    stopping = "stopping"
+    stopped = "stopped"
+
+
+Lifecycle = ExecutionLifecycle
```
tmi: this is an odd line -- is `Lifecycle` on its own used in this file? We ideally want just one name/concept, so best remove this line?

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+
+
+@dataclass(frozen=True, eq=True, slots=True)
+class RestartPolicy:
+    """A finite restart budget for one managed long-lived thread."""
+
+    def __init__(self, max_restarts: int = 0) -> None:
+        if max_restarts < 0:
+            raise ValueError("max_restarts must be non-negative")
+        object.__setattr__(self, "max_restarts", max_restarts)
```
tmi: use the __post_init__ method instead for this validation-like code so that you dont need to do the setattr thing

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+@dataclass(frozen=True, eq=True, slots=True)
+class ComponentStatus:
+    ready: bool
+    health: str = "unknown"
+    detail: str = ""
+
+
+@dataclass(frozen=True, eq=True, slots=True)
+class MonitoredFailure:
+    pool_name: ConcurrentPools | str
```
tmi: why would we have the `| str` here? Unions make sense when we do case matching, ie, the class itself encodes behaviour. But if it is just for sake of "caller convenience" and both cases encode the same information, lets not do that

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+    threads: Mapping[str, ThreadStatus]
+    monitored_failures: tuple[MonitoredFailure, ...]
+    unregistered_threads: tuple[str, ...]
+
+
+_worker_context = threading.local()
+
+
+def _freeze_mapping(values: Mapping[Any, Any]) -> Mapping[Any, Any]:
+    return MappingProxyType(dict(values))
```
tmi: put this function to utility/structural.py, make it public

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+        with self._lock:
+            executor = self._executor
+            if executor is None or not self._accepting:
+                self._permits.release()
+                raise SubmissionRejected(f"pool is not accepting submissions: {self.pool_name.value}")
+            self._submitted += 1
+            self._pending += 1
+
+        def wrapped() -> T:
+            _worker_context.pool_name = self.pool_name
```
tmi, line 237: why do we set pool_name back and forth? Why dont we set it once during initialization or first submit? The pool name is immutable for the whole life of the pool

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+            raise ValueError("failure_history_size must be positive")
+        self._lock = threading.RLock()
+        self._lifecycle = ExecutionLifecycle.new
+        self._pools: dict[ConcurrentPools, ManagedPool] = {}
+        self._threads: dict[str, _ThreadRecord] = {}
+        self._ready_stages: set[int] = set()
+        self._monitored_failures: deque[MonitoredFailure] = deque(maxlen=failure_history_size)
+
+    @staticmethod
+    def _pool_name(pool_name: ConcurrentPools | str) -> ConcurrentPools:
```
tmi: this method should not exist -- do not operate with the `str` representation at all, eg in the register_pool method, use always the enum members

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+        try:
+            return ConcurrentPools(pool_name)
+        except ValueError as error:
+            raise RegistrationError(f"unknown pool: {pool_name}") from error
+
+    def has_pool(self, pool_name: ConcurrentPools | str) -> bool:
```
tmi: this method should not be needed, remove it

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+        with self._lock:
+            if self._lifecycle not in (ExecutionLifecycle.new, ExecutionLifecycle.starting):
+                raise LifecycleError("pools can only be registered before or during startup")
+            if normalized in self._pools:
+                raise RegistrationError(f"duplicate pool: {normalized.value}")
+            self._pools[normalized] = ManagedPool(normalized, max_workers, max_pending, stage)
+
+    def register_thread(
+        self,
+        thread_name: str,
```
tmi: the `thread_name` in this class should not be `str`, but a member of the ConcurrentThreads enum

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+                record.uncaught_exception = repr(error)
+                record.traceback = traceback.format_exc()
+            finally:
+                record.stopped_at = time.time()
+
+        record.thread = threading.Thread(name=f"fiab-{record.thread_name}", target=run)
+        record.thread.start()
+
+    @staticmethod
+    def _provider_ready(provider: object) -> bool:
```
tmi: we need to improve on this. We will already make a change that the provider object must be a pydantic class, but it appears we should additionally require it defines a boolean is_ready() method. We cant have heuristic like this here

backend/src/forecastbox/utility/concurrency/manager.py
```diff
@@ -0,0 +1,674 @@
...
+            managed_thread = any(record.thread is current for record in self._threads.values())
+            if (
+                managed_thread
+                and self._lifecycle in (ExecutionLifecycle.starting, ExecutionLifecycle.stopping)
+                and pool.stage in self._ready_stages
+            ):
+                return
+            raise SubmissionRejected(f"manager is not accepting submissions: {self._lifecycle.value}")
+
+    def _pool(self, pool_name: ConcurrentPools | str) -> ManagedPool:
```
tmi: dont accept a string here, only enum instances

backend/src/forecastbox/utility/dispatcher.py
```diff
@@ -0,0 +1,340 @@
...
+
+class DispatcherClosed(DispatcherError):
+    """Raised when event intake has been permanently closed."""
+
+
+class DispatcherRegistrationError(DispatcherError):
+    """Raised for invalid dispatcher registrations."""
+
+
+def _freeze_payload(value: object) -> object:
```
tmi: move to utils/structural, rename to `freeze_recursively`

PR #631, opened 2026-08-06
Feat/glyphs to run detail

tmi:
Adding glyph values to RunDetail

Allowing /expand to be run in validateOnly mode

Adding integration tests that check /expand-before-submit ~ /expand-after-submit, and that /expand after submit has the correct implicit glyph values


tmi:
cc @liefra: have a look at the new field in the in the RunDetailResponse, you can use that to pass to the /expand after a submit. You can have a look at the integration test changes, thats basically the flow you are interested in -- retrieve the original builder, replace the local glyphs with what was given in the RunDetailResponse, call /expand with validateOnly=False, use the returned resolved configuration options

PR #633, opened 2026-08-07
Feat/artifact notifications

tmi:
artifact: emit ClientNotification when a download completes

Adds domain/artifact/events.py with ArtifactDownloadedEvent, implementing
the ClientNotificationSource protocol, emitted from manager.py once a
downloaded artifact becomes locally available. Progress reports are
intentionally not surfaced this way, clients keep polling for those.

Adds test utilities connect_notification_websocket and
wait_next_notification to tests/integration/utils.py, and refactors
test_notification.py to use them. Extends test_model.py to assert that
matching artifactDownloaded notifications arrive for all downloaded
models, and that their refreshRoutes resolve to a working endpoint.

### DIFF
--- a/backend/src/forecastbox/domain/artifact/manager.py
+++ b/backend/src/forecastbox/domain/artifact/manager.py
@@ -30,7 +30,7 @@ from pyrsistent.typing import PMap, PSet
 
 from forecastbox.domain.artifact.base import ArtifactCatalog, CompositeArtifactId, MlModelDetail, MlModelOverview
 from forecastbox.domain.artifact.catalog import get_artifacts_catalog
-from forecastbox.domain.artifact.events import ArtifactDownloadedEvent
+from forecastbox.domain.artifact.events import ArtifactDownloadFinishedEvent
 from forecastbox.domain.artifact.io import delete_artifact, download_artifact, list_storage
 from forecastbox.utility import tunnel
 from forecastbox.utility.concurrency.synchronization import timed_acquire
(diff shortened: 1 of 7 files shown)


backend/src/forecastbox/domain/artifact/manager.py
```diff
@@ -149,6 +151,12 @@ def progress_callback(progress: int) -> None:
     except Exception as e:
         logger.exception(f"artifact download failed for {composite_id}: {repr(e)}")
         report_artifact_download_progress(composite_id, failure=repr(e))
+        return
```
tmi: TODO we need to distinguish failure here, send a different notification presumably

backend/tests/integration/test_notification.py
```diff
@@ -11,40 +11,29 @@
...
-            # a notification from a different domain/test running concurrently -- ignore it
-
-        assert notification is not None, f"did not receive a matching test notification for identifier {identifier}"
-        assert notification["sourceDomainName"] == "notification"
-        assert notification["sourceDomainEvent"] == "testNotification"
-        assert identifier in notification["text"]
+        notification = wait_next_notification(websocket, "notification", "testNotification", total_timeout=15)
+        while notification.context.get("identifier") != identifier:
+            # a testNotification from a different, concurrently running test -- ignore it
+            notification = wait_next_notification(websocket, "notification", "testNotification", total_timeout=15)
```
tmi: TODO the 15 should be progressively decreased

tmi:
cc @liefra -- here comes a first real notification, upon artifact download completion. I did it so that there is context with the artifact id, and there is both detail url and the refresh urls. Not actually urls, but the path suffixes (like "api/v1/artifacts/list_models", since the frontend knows the host-port better)

PR #640, opened 2026-08-10
run: persist resolved configuration options in compiler runtime context

tmi:
Add a resolution field to CompilerRuntimeContext, populated during background compilation with the per-block configuration options that were resolved via glyph substitution. compile_builder now returns this mapping alongside the existing compilation outputs.

Replace RunDetail.glyph_values (and the corresponding route response field) with resolution, sourced directly from the persisted runtime context.

Rework the integration test that inspected glyph_values to instead verify the resolution field.

### DIFF
--- a/backend/src/forecastbox/domain/run/background.py
+++ b/backend/src/forecastbox/domain/run/background.py
@@ -95,10 +95,10 @@ def execute_background(
         relevant_glyphs_and_values = expand_glyph_values(all_glyphs_raw, roots=referenced_glyph_names)
         used_glyphs = {k: all_glyphs_raw[k] for k in relevant_glyphs_and_values.keys() if k not in PINNED_INTRINSIC_KEYS}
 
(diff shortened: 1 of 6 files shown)


tmi:
cc @liefra this replaces the glyph_values with resolved configuration options in the RunDetail response. Meaning you don't need to call `/expand` anymore, you simply gather all the configuration options from this response when displaying, and use them as they are

a caveat: we _only_ include configuration options that included some glyph. Meaning when there were any completely static ones, use what was in the builder. Thats a bit of an oddity, but I think it mirrors how the `/expand` endpoint itself is behaving with its `resolved_configuration_options` field.

a caveat: this is not backwards incompatible on persistence level, we simply add a new field into a json that we persist with a defaut defined. But for jobs that were submitted before this PR as well as for jobs that were not compiled yet, you will simply get a None in this field. Hence best use it opportunistically -- build the model exactly as you did before, and _if_ there is this field, use it to replace the provided configuration option ids

PR #647, opened 2026-08-19
[plugin.ecmwf] add dummy input anemoi

tmi:
Adding anemoi dummy input for case where the user does not have a valid mars key or opendata is timeouting too much


tmi:
demonstrated on a single run with o48 global + selects + zarr sink, worked fine

also tried it with grib sink, but that failed with
```
  File "/tmp/cascade_worker_venv_1e59pe02/lib/python3.12/site-packages/earthkit/data/core/plugins.py", line 162, in find_plugin
    raise NameError(f"Cannot find {loader.kind} '{name}' (values are: {candidates})")
NameError: Cannot find encoder 'dict' (values are: internal-pass-through)
```
, not sure why. I gave it path `/tmp/o1` -- is the templating string or suffix or anything required? Can we add a validation?

PR #652, opened 2026-08-20
Feat/plugin notifications

tmi:
Add plugin notifications for install, update, settings applied, uninstalled


tmi:
cc @liefra -- the remaining notifications for plugins

PR #653, opened 2026-08-21
plugins: activate editable installs in-process, no restart needed

tmi:
An editable (-e) plugin install writes a new .pth file into site-packages instead of placing files under an existing sys.path entry. .pth files are only read by the site module at interpreter start-up, so a long-running backend process never picks up the new source location and the plugin stays unimportable until restarted.

Add forecastbox.utility.pth_activation, a stdlib-only module that replays what site.addsitedir/site.addpackage would do at start-up, but scoped to only the .pth file(s) a specific install just created. This handles both the classic setuptools path-based editable mode and PEP 660 finder-based editable installs (hatchling, pdm, flit, ...), without assuming a particular source layout.

Wire this into install_plugin_compatibly, right after a successful real install: snapshot .pth filenames before/after, activate any new ones, and invalidate import caches. Activation only runs in-process when the target python is verified to be the interpreter actually executing this code (captured once at import time), so it is a no-op for callers that target a different interpreter (e.g. tests/adhoc's scratch venv, which only reassigns sys.executable on the outer process).

Add unit tests for the new module, and a new adhoc scenario that performs a fresh editable install and import in a subprocess running under the scratch venv's own interpreter, genuinely exercising the fix end-to-end. The integration tests are unchanged  -- they currently assume the plugin is pre-installed and don't exercise the runtime pip-install path.

### DIFF
--- a/backend/src/forecastbox/domain/plugin/compatibility.py
+++ b/backend/src/forecastbox/domain/plugin/compatibility.py
@@ -45,8 +45,12 @@ Algorithm for ``install_plugin_compatibly``
    requirements, and the requested plugin requirement.
 6. Only if the dry run succeeds, run the identical command for real (differing only by the
    absence of ``--dry-run``).
-7. Run ``uv pip check`` again as a post-install check.
-8. Return the parsed installed-version mapping (from the real install's output only) for the
+7. If the real install's target interpreter is the one actually running this code (see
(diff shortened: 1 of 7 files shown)


tmi:
Tested by hand with `just dev` with a venv free of fiab-plugin-test, triggered install with `-e` pip url in config.toml, import now works right away

PR #654, opened 2026-08-21
[docs] describe local launching for devs

tmi:


tmi:
@corentincarton -- does this look useful? I added explanation on the difference between `just dev` and `fiab.sh`, in a standalone doc as that is valuable on its own, and added references between that and the existing `plugins.md`

PR #655, opened 2026-08-24
[deployment] add slim docker image build

tmi:
putting to `v2` folder for now to differentiate from the original ewc deploy

for start just `build` and `run` commands, tested both

will add publish and test later


tmi:
cc @HCookie -- the `fiab.sh`-only docker image, without publish or test yet

PR #662, opened 2026-08-25
[backend] remove outdated git dependency

tmi:
we dont need git anymore, because fiab-core is now equipped with automatically updated fallback versions


tmi:
cc @HCookie -- it is actually even better, we should not need the git anymore at all; thx again for the report

PR #665, opened 2026-08-25
feat: MARS keys in model metadata

HCookie:
### Description
At request of @jinmannwong , adding mars metadata to anemoi models

Includes a change to allow for arbitrary extra metadata, and adapts the existing checkpoints to include stream, type & class.

### Contributor Declaration

### DIFF
--- a/backend/packages/fiab-core/src/fiab_core/artifacts.py
+++ b/backend/packages/fiab-core/src/fiab_core/artifacts.py
@@ -109,8 +109,6 @@ class AnemoiCheckpoint(FiabCoreBaseModel):
     input_qube: dict[str, Any] = Field(description="Json Dump of the input qube structure, including variables, levels")
     output_qube: dict[str, Any] = Field(description="Json Dump of the output qube structure, including variables, levels, but not step")
 
-    extra_metadata: dict[str, list[Any]] = Field(default_factory=dict, description="Additional metadata, i.e. MARS keys")
-
     configuration: AnemoiCheckpointConfiguration = Field(
         default_factory=AnemoiCheckpointConfiguration,
         description="Additional configuration for the checkpoint such as pre and post processors and control options",
--- a/backend/packages/fiab-core/src/fiab_core/artifacts.py
+++ b/backend/packages/fiab-core/src/fiab_core/artifacts.py
@@ -109,6 +109,8 @@ class AnemoiCheckpoint(FiabCoreBaseModel):
     input_qube: dict[str, Any] = Field(description="Json Dump of the input qube structure, including variables, levels")
     output_qube: dict[str, Any] = Field(description="Json Dump of the output qube structure, including variables, levels, but not step")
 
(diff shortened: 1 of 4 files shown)


backend/packages/fiab-core/src/fiab_core/artifacts.py
```diff
@@ -109,6 +109,8 @@ class AnemoiCheckpoint(FiabCoreBaseModel):
     input_qube: dict[str, Any] = Field(description="Json Dump of the input qube structure, including variables, levels")
     output_qube: dict[str, Any] = Field(description="Json Dump of the output qube structure, including variables, levels, but not step")
 
+    extra_metadata: dict[str, list[Any]] = Field(default_factory=dict, description="Additional metadata, i.e. MARS keys")
```
tmi: wouldn't a better name be `extra_qube_expansion`, because we always use it as such? Because "Additional/extra metadata" suggests like "put anything you want here", very generic. But we mandate the `expand_qube` will succeed with that. Even the `list[Any]` could perhaps be improved? And is the annotation even correct? Because we use in the _expand_qubed, which accepts `dict[str, Any]`, not `list[Any]`, right? Or do I misunderstand it? And the artifacts.json gets values in the `children` field, not the `extra_metadata`, so... I'm confused :) It's not a big deal if it works and can be merged as such, it just reads difficult to me...

tmi (approved):

backend/packages/fiab-core/src/fiab_core/artifacts.py
HCookie: I've removed it, as it is possible to just modify the output qube definition in the `artifacts.json` rather than need a special key. Thereby keeping backwards compat

backend/packages/fiab-plugin-ecmwf/scripts/get_metadata_from_anemoi.py
```diff
@@ -40,6 +43,8 @@ def get_qubes(checkpoint_path: str) -> dict[str, Any]:
     in_qube = _expansion_qube(in_variables, variables_metadata, model_step, 6).remove_by_key("step")
     out_qube = _expansion_qube(out_variables, variables_metadata, model_step, 6).remove_by_key("step")
 
+    out_qube = out_qube | Qube.from_datacube({"class": "ai", "type": "fc", "stream": "oper"})
```
jinmannwong: Why are the values for `class`, `type` and `stream` hardcoded here? Do we not use the values in the artifacts.json?

HCookie:
```suggestion
```
Whoops, that was a debug line

jinmannwong (approved):

backend/packages/fiab-plugin-ecmwf/scripts/get_metadata_from_anemoi.py
HCookie: I don't think it was.....

PR #667, opened 2026-08-25
feat(plugins): Geographical blocks

### LINKED ISSUE #581: Regional interpolation subset blocks
Add dedicated blocks for regional subset/interpolation.

---
_Created by an AI agent from the FIAB Miro planning board._

HCookie:
### Description
Add regridding & area subselect blocks

Closes #581 

## Example
image

### Contributor Declaration


HCookie:
@tmi What are your thoughts on the grid type I attempt to introduce here?

backend/packages/fiab-core/src/fiab_core/types/definitions.py
```diff
@@ -73,11 +73,18 @@ def serialize(self) -> str:
...
+        """If real is True, only allow real numbers (no NaN or inf)."""
+        self.real = real
+
     def validate_convert(self, value: Any) -> float:
         if not isinstance(value, str):
             raise NotStringInput(f"Expected string, got {type(value).__name__}")
         try:
-            return float(value)
+            result = float(value)
+            if self.real and (result != result or result in (float("inf"), float("-inf"))):
```
tmi: nit: use the `not (math.isnan(result) or math.isinf(result))`, I mean this works too but its quirky

backend/packages/fiab-core/src/fiab_core/tools/blocks.py
```diff
@@ -211,6 +211,11 @@ def config_as_geodomain(self, key: str | ConfigurationOptionId) -> GeoDomainWrap
         # NOTE we ignore types as we trust parser -- but its fragile!
         return GeoDomainWrapper(raw_value)  # ty:ignore[invalid-argument-type]
 
+    def config_as_gridspec(self, key: str | ConfigurationOptionId) -> GridWrapper:
+        option_id, option = self._get_configuration_option(key)
+        raw_value = self._get_raw_value(option_id)
```
tmi, line 233: you best check the isinstance here, or put the isinstance into the GridWrapper validator -- unlike pydantic, dataclasses dont check on their own at runtime

backend/packages/fiab-core/src/fiab_core/types/definitions.py
```diff
@@ -310,6 +317,72 @@ def serialize(self) -> str:
...
+
+    def serialize(self) -> str:
+        return "tuple-float-grid"
+
+
+class GridType(UnionType):
+    """An alias for a union over named grid and tuple of floats representing grid resolution."""
+
+    def __init__(self) -> None:
+        super().__init__([NamedGridType(), TupleFloatGridType(), LatLonGridType()])
```
tmi: do you need the other three to be their own types? You could instead have GridType as a new type which does not inherit any existing type, and handles the case-match itself, instead of relying on the union first-match-wins algorithm. Perhaps the question is how often will they appear on their own (ie, a block requiring exactly one subtype) and how often they would appear in blocks as the union GridType only

happy with the subtypes being explicit -- but note we need to communicate to frank as well to handle them, parse them, etc. Note you best define serialze for the LatLonGrid as well. I would probably serialize them as `grid`, `grid-latlon1`, `grid-gaussian`, `grid-latlon2`...

backend/packages/fiab-core/src/fiab_core/tools/convert.py
```diff
@@ -47,3 +47,22 @@ def with_bbox_mars(self) -> Self:
...
+
+
+GridRuntimeType = str | list[float]
+
+
+@dataclass
+class GridWrapper:
+    value: GridRuntimeType
+
+    def as_grid_spec(self) -> dict[Literal["grid"], str | list[float]]:
```
tmi: do you even need this wrapper? Why would the `config_as_gridspec` return directly the invocation of this `as_grid_spec()`?

HCookie: I think I named the config one wrong, this might be expanded as I build out the MIR regridding stuff

backend/packages/fiab-core/src/fiab_core/types/definitions.py
HCookie: I was following the convention from geodomain, I can rework it to be a GridType completely?

HCookie: I could imagine tho the subtypes being used within restrictions, once we have knowledge of the grid the input is, we can restrict the output...

tmi: > I was following the convention from geodomain, I can rework it to be a GridType completely?

with the geodomain, it was easier to express the union of existing types without any custom validation imo, and I though that at that time we explicitly discussed that some blocks will use only subset of options. Either way of doing this is fine, it's just that generally I would start simple (one type) and split only when needed.

HCookie:
Runtime implementation added, 
It uses `mir` to handle both area cutout and regridding in one block. Due to no optional support, I introduced an `auto` grid, which translates to, do nothing at all. 
One issue identified was that after doing a cutout ekp kinda breaks as it can't regrid back to a regular grid, so for now ensure that if you cutout, also regrid in the same step.

tmi (approved):

PR #677, opened 2026-08-27
feat/lensProxy

tmi:
Changes the lens url -- instead of returning the underlying port, we return an URL that points to backend's new route for proxying, and the backend then proxies itself, additionally applying auth

Minor tech debt / refactor: we centralize the /api/v1 prefix, it now comes from config into all places

*Release Notes*: enables Lens functionality in cloud/containerized deployments. Makes Lens require (any) authentication

### DIFF
--- a/backend/src/forecastbox/domain/lens/proxy.py
+++ b/backend/src/forecastbox/domain/lens/proxy.py
@@ -1,0 +1,229 @@
+# (C) Copyright 2024- ECMWF.
+#
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+#
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+"""Lens proxy -- exposing internally launched lens processes via the backend's own port.
+
+A lens is an external process (e.g. skinnyWMS) that the backend launches and which
+binds to 127.0.0.1:<port> on a dynamically chosen port. In containerized deployments
+we cannot expose those dynamic ports to the outside world, so instead every lens is
+reachable through a fixed path on the backend's own port:
+
+    /api/v1/lens/proxy/<lens_instance_id>/<upstream path>[?<query>]
+
+This module implements a streaming HTTP reverse proxy for that path. The following is
+the contract between the backend proxy, the clients that call it, and the lens
+processes being proxied. It is authoritative; new clients (beyond the current
+frontend, e.g. the CLI) and new lens types must adhere to it.
+
+Supported traffic
+------------------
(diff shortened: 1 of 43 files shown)


backend/src/forecastbox/domain/lens/proxy.py
```diff
@@ -0,0 +1,229 @@
...
+import httpx
+from fastapi import HTTPException, Request
+from fastapi.responses import StreamingResponse
+from starlette.datastructures import Headers
+
+from forecastbox.domain.lens.manager import LensInstanceId, get_status
+
+logger = logging.getLogger(__name__)
+
+PREFIX = "/api/v1/lens/proxy"
```
tmi: TODO: duplicated, we must unify with the routes prefix

backend/src/forecastbox/domain/lens/proxy.py
```diff
@@ -0,0 +1,229 @@
...
+    not `running`, or if it has a number of ports other than exactly one (an
+    invariant violation for the currently supported lens types).
+    """
+    detail = get_status(lens_instance_id)
+    if detail.status != "running":
+        raise RuntimeError(f"Lens instance {lens_instance_id!r} is not running (status={detail.status!r})")
+    if len(detail.ports) != 1:
+        logger.error(f"Lens instance {lens_instance_id!r} does not expose exactly one port: {detail.ports!r}")
+        raise RuntimeError(f"Lens instance {lens_instance_id!r} does not expose exactly one port")
+    return next(iter(detail.ports))
```
tmi: why not [0]?

backend/src/forecastbox/domain/lens/proxy.py
```diff
@@ -0,0 +1,229 @@
...
+
+    Raises HTTPException(404) if the lens instance is unknown or not running, and
+    HTTPException(502/503) if the upstream connection fails.
+    """
+    try:
+        port = _resolve_port(lens_instance_id)
+    except KeyError:
+        raise HTTPException(status_code=404, detail=f"Lens instance {lens_instance_id!r} not found")
+    except RuntimeError as e:
+        raise HTTPException(status_code=404, detail=str(e))
```
tmi: TODO http exceptions should live in routes, not here

backend/src/forecastbox/domain/lens/proxy.py
```diff
@@ -0,0 +1,229 @@
...
+        headers=headers,
+        content=request.stream(),
+    )
+
+    try:
+        upstream_response = await client.send(upstream_request, stream=True)
+    except httpx.ConnectError:
+        raise HTTPException(status_code=502, detail=f"Lens instance {lens_instance_id!r} is unreachable")
+    except httpx.TimeoutException:
+        raise HTTPException(status_code=504, detail=f"Lens instance {lens_instance_id!r} timed out")
```
tmi, line 213: TODO add a comment here that these http exceptions are ok

PR #679, opened 2026-08-28
[backend] fix(lens): do not frame bodyless proxy requests as chunked

liefra:
Follow-up to #677: the lens proxy silently truncates WMS responses.

## Issue

Repeated GetMap through `/api/v1/lens/proxy/<id>/...` returns short bodies as `200 OK`. 40 sequential requests over one connection, proxied vs. the lens's own port:

| layer | direct | proxied |
|---|---|---|
| 2t | 100% | 72.5% |
| msl | 100% | 60.0% |
| 2d | 100% | 55.0% |
| 10v | 100% | 57.5% |

Six of seven layers across two output directories, so not dataset-specific. In the geo viewer the partial PNG fails to decode and the tile renders as an error.

## Cause

`forward` passed `content=request.stream()` for every request. WMS requests are bodyless GETs, but the stream has no known length, so httpx framed them `Transfer-Encoding: chunked`. gunicorn's sync worker never drains a request body on a GET — it answers and closes the socket with that body unread, the kernel sends RST instead of FIN, and the response body still being streamed back is cut off. Server-side it appears as `httpx.ReadError` out of `body_iterator`.

## Fix

Only stream a body when the request has one. Requests that do carry a body are unaffected.

## Verification

All failing layers now 100%, bodies byte-exact, throughput unchanged (~2% proxy overhead). New unit test fails on current `main`, passes with the fix.

### Contributor Declaration

### DIFF
--- a/backend/src/forecastbox/domain/lens/proxy.py
+++ b/backend/src/forecastbox/domain/lens/proxy.py
@@ -176,6 +176,22 @@ def _build_upstream_headers(request: Request, port: int, lens_instance_id: LensI
     return headers
 
 
+def _has_body(request: Request) -> bool:
+    """Does this request carry a body worth streaming upstream?
+
+    Passing a stream for a bodyless GET makes httpx frame it as
+    Transfer-Encoding: chunked. gunicorn's sync worker never drains that body,
+    so it closes the socket with request data unread, the kernel answers RST,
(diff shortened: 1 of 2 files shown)


tmi:
I had to add one extra ty:ignore for some reason, pushing anew in https://github.com/ecmwf/forecast-in-a-box/pull/680

PR #695, opened 2026-09-03
[plugin/ecmwf] add missign dep to operational src

tmi:
the operational forecast source block requires ecmwf-opendata dependency when it is run with ecmwf-openadata source

### DIFF
--- a/backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
+++ b/backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
@@ -160,9 +160,9 @@ class OperationalForecastSource(Source):
 
         source = block.config_as_str(SOURCE)
         if source == "ecmwf-opendata":
-            metadata = {"environment": ["earthkit-data[ecmwf-opendata]"]} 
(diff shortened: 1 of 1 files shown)


backend/packages/fiab-plugin-ecmwf/src/fiab_plugin_ecmwf/blocks.py
```diff
@@ -158,6 +158,9 @@ def compile(
         date = basetime.strftime("%Y%m%d")
         time = self._convert_time(basetime.time().hour)
 
+        source = block.config_as_str(SOURCE)
+        metadata = {"environment": ["ecmwf-opendata"]} if source == "ecmwf-opendata" else {}
```
jinmannwong:
```suggestion
        if source == "ecmwf-opendata":
            metadata = {"environment": ["earthkit-data[ecmwf-opendata]"]} 
        elif source == "mars":
            metadata = {"environment": ["earthkit-data[mars]"]} 
        else:
            metadata = {}
```

jinmannwong: Thanks for catching this, though I don't understand how I've not encountered this issue with the missing dependency before. Just added a suggested change to also cover the case where the source is mars.

PR #702, opened 2026-09-03
[deployment] dev docker

tmi:
splits the docker `just build` command into buildRelease and buildDev

factors out the common platform docker file, and the two release/dev specifics docker files

makes the fiab.sh in the buildRelease to be pulled at runtime instead of copied from the repo

creates the dockerfile-dev to be based on `just dev` and `uv sync`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -11,7 +11,6 @@ backend/MANIFEST.in
 **/node_modules
 
 # Build artifacts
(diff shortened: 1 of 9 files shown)


tmi:
cc @HCookie this should make it easier to build a docker from dev. It's a bit hacky at places, please report any inconvenience faced

PR #704, opened 2026-09-03
fiab-core: add TimeDeltaType for ISO 8601 durations

tmi:
Introduce a new FableType for datetime.timedelta values, serializing as 'timedelta'. Parsing follows ISO 8601 duration format (weeks, days, hours, minutes, seconds), with the leading 'P' designator optional on input (e.g. 'T12H' or '3DT1M' are accepted alongside 'PT12H'/'P3DT1M'). Calendar-dependent components (years, months) are not supported since they do not map to a fixed-length duration.

Registered the new atomic type in the parser and added unit tests, including combinations with list[timedelta] and union types (with string and datetime members, in both orderings).

### DIFF
--- a/backend/packages/fiab-core/src/fiab_core/types/__init__.py
+++ b/backend/packages/fiab-core/src/fiab_core/types/__init__.py
@@ -11,15 +11,17 @@
 FableType: Type system for Forecast As BLock Expression (Fable) configuration values.
 
 Provides parsing, validation, and conversion for a small set of type expressions:
-- str, int, float, date, datetime (atomic types)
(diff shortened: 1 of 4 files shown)


backend/packages/fiab-core/src/fiab_core/types/__init__.py
```diff
@@ -12,6 +12,7 @@
 
 Provides parsing, validation, and conversion for a small set of type expressions:
 - str, int, float, date, datetime (atomic types)
+- timedelta (ISO 8601 duration, converted to datetime.timedelta)
```
tmi: better

PR #705, opened 2026-09-04
fix: Switch MARS keys to extra_output_keys 

HCookie:
### Description
Switch MARS keys to supplementary_metadata, and set as scalar fields

### Contributor Declaration


tmi (approved):
approving but please don't name it 'supplementary_metadata', thats a combination of two generic terms which convey no concrete meaning. And we already have "metadata", how does it differ from this? Furthermore, this is not metadata because the action utilizes it for execution, this is very much data. Metadata is like author or upload date or whatever. Can we name it "initial_conditions_extra_keys" or something like that? If you think it is too specific, I don't mind nesting it in something like "additional_execution_data" or "runtime_overrides" etc

(I understand that from the original ML PoV, checkpoint is data:weights + metadata:everythingElse, but the everythingElse is too big and important and diverse by now already that we need to finegrain further)

PR #714, opened 2026-09-09
[plugin.ecmwf] fix dep declaration

tmi:
replacing earthkit-data[smth] with direct dependency due to install time quirk


tmi:
tested with docker v2, just buildDev -- the single IFS snapshot + wms now (finally) work e2e :) 

cc @HCookie @jinmannwong -- please don't use `earthkit-data` or `earthkit-workflows` as dependencies for Actions. Ideally we don't want to use the `[...]` at all, because its generally unclear how to resolve conflicts there (say one package says `pgk1[A]` and another `pgk1[B]` -- is _always_ `[A, B]` the correct result? And can we actually _reliably_ _retrieve_ that information via pip freeze? I will have a look at it later, but until then please be careful :)

tmi:
ok @HCookie @jinmannwong what I said _may_ have been irrelevant/incorrect, because there was a much more obvious bug, see my most recent commit in this branch (which I append to my already long list of "use enums, not strings; and use assert_never" evidences)

but I'm going to leave the indirection/lookup in place anyway, because my discomfort about `[]` deps and how to resolve them in case of conflict remains...

PR #717, opened 2026-09-10
[backend.run] fix the run delete

tmi:
1/ do not mandate cascade_job_id to be not None -- that was a bug which caused unsubmitted jobs to linger forever

2/ when actually submitting a cascade job, check whether the underlying entity is soft deleted -- and if so, issue a (delayed) cascade deletion request right away

3/ do not mandate the delete's attempt count to match the latest, we do allow to delete previous runs from db/cascade as well

### DIFF
--- a/backend/src/forecastbox/domain/run/background.py
+++ b/backend/src/forecastbox/domain/run/background.py
@@ -33,10 +33,11 @@ from forecastbox.domain.glyphs.resolution import (
     merge_glyph_values,
 )
 from forecastbox.domain.run import db
-from forecastbox.domain.run.cascade import execute_cascade
+from forecastbox.domain.run.cascade import delete_cascade_job, execute_cascade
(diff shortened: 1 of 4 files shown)


tmi:
cc @HCookie this fixes the bug you reported, thx again. There was actually more rot than just that

cc @liefra there was previously an explicit requirement that only the latest attempt_count of a job was accepted for deletion, on the grounds of "preventing a race condition". I can't see a reason for that now, hence I lifted the restriction, and any attempt_count of a run can now be deleted. But the attempt_count is still mandatory, ie, there is no "delete latest attempt count of a run" (that _would_ threaten a race condition!). As before, the deletion is soft, and triggers a request to cascade itself to delete the job, possibly delayed in case it was not submitted yet

We actually have a gap in cascade that we currently only allow outputs deletion, but I'll fix that later

PR #725, opened 2026-09-16
Bug/scheduler init

tmi:
backend: gate scheduler start on plugin catalog via explicit barriers

    Fix a startup race where the scheduler could start before the plugin
    catalog finished loading. Introduces two one-shot barrier futures in
    entrypoint/initializers.py (ArtifactsCatalogInitialized,
    PluginsCatalogInitialized) so the previously coupled artifact-manager
    plugin-catalog initializer step can be split into two independent
    steps, with the scheduler gated on the plugin catalog barrier instead.

    Barriers forward the same outcome (success, failure, or cancellation)
    as the step they are attached to, so a failure earlier in the chain is
    observed by dependents instead of silently treated as success.

    Supporting changes:
    - ExecutionManager.submit_after now returns a Future mirroring the
      eventual outcome of the submitted task (or the dependency's failure),
      instead of being fire-and-forget.
    - domain.plugin.submit.submit_load_all returns that future.
    - domain.experiment.scheduling.background.start_scheduler accepts an
      optional blocking future to gate the actual start on, defensively
      logging and skipping the start if that future failed, without
      raising further.
    - tests/unit/conftest.py's inline submit_after stub updated to match
      the new return-a-future contract.


tmi:
cc @liefra -- this replaces the "plugin not found" bug by "Gateway is not running" bug :) Ie, we are progressing but not there yet

PR #733, opened 2026-09-21
feat: support file:// URLs for checkpoint artifacts

HCookie:
## Summary

Adds support for `file://` URLs when downloading checkpoint artifacts locally. When an artifact's `url` starts with `file://`, the file is copied directly from the local path instead of going through the HTTP download path.

## Changes

- `backend/src/forecastbox/domain/artifact/io.py`: in `_download_artifact_local`, short-circuit when `checkpoint.url` is a `file://` URL and `shutil.copy` the resolved local path to the artifact destination.

## Motivation

This makes it possible to register and use checkpoints that already exist on the local filesystem (e.g. during development or on air-gapped deployments) without needing to serve them over HTTP.

## Notes

Extracted from downstream work in a fork (original author: Vegard Bønes).


tmi (approved):
makes sense, just did minor improvements in https://github.com/ecmwf/forecast-in-a-box/pull/738

PR #734, opened 2026-09-21
fix: align get_metadata_from_anemoi.py output with common/specific artifact schema

HCookie:
## Summary

The `get_metadata_from_anemoi.py` helper script emits an artifact entry that no longer matches the `fiab_core.artifacts` Pydantic schema. This PR brings the script back in sync.

## Problems fixed

1. **Schema shape mismatch** — the script emits a single `store_info` block, but `ArtifactsProvider` parses `common` (`CommonArtifactMetadata`) and `specific` (`AnemoiCheckpoint`) blocks. Fields are now split accordingly:
   - `common`: `url`, `display_name`, `display_author`, `display_description`, `comment`, `disk_size_bytes`, `supported_platforms`
   - `specific`: `minimum_gpu_memory_mib`, `pip_package_constraints`, `input_characteristics`, `input_qube`, `output_qube`, `extra_output_keys`, `timestep`
2. **`extra_output_keys` type mismatch** — the schema declares `dict[str, str]`, but the script emitted list values (`{"class": ["ai"], ...}`). Changed to scalar values (`{"class": "ai", ...}`).
3. **Removed stale `input_options: {}`** — this key is not a field of `AnemoiCheckpoint` (it lives under `configuration`), so emitting it at the top level was incorrect.
4. **Editable/git installs of dependencies** — `module_versions` can contain a DotDict (with `version`/`source`) instead of a plain string for non-registry installs, which produced malformed pip constraints. Now unwrapped to the `.version` attribute.
5. **Dependency bumps** — added `earthkit-workflows>=0.9.0` and bumped `earthkit-workflows-anemoi>=0.7.2` for the script's inline dependency block.

## Changes

- `backend/packages/fiab-plugin-ecmwf/scripts/get_metadata_from_anemoi.py`

## Notes

Extracted from downstream work in a fork (original author: Vegard Bønes).


tmi (approved):
legit -- but I wonder, why do you create raw dicts and then json dumps? You could instead use the `fiab-core` pydantic objects. That would have prevented this bug from appearing in `main` because `ty` would have stopped the offending PR

tmi (approved):

PR #738, opened 2026-09-22
[backend] artifacts io minor improvements

tmi:
[backend] artifacts io minor improvements

    1. wraps the local file copy with a try catch to not crash the backend
    2. adds the progress callback invocation to uphold the contract
    3. instead does a hardlink in case the same filesystem, for better perf/disk


tmi:
cc @vegardb (cant assign you as a reviewer...)

